PR #1, opened 2018-02-22
Add docker files

luisremis:
Adding the docker file to build a docker image of VDMS, and a README.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,24 @@
+# VDMS Docker
+
+Here are the instructions to build a docker image of the VDMS Server.
+
+The build command is: 
+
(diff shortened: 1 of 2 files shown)


docker/README.md
```diff
@@ -0,0 +1,21 @@
+# VDMS Docker
+
+The build command is:
```
crstrong: Do we need to specify that this is for building the server only? I'd like to think people are smart enough to figure that out, but ... 

If we mention it somewhere else (like in the main README), that's also probably fine

luisremis: I added that to the README here, just in case.

vishakha041:
Wonderful

PR #4, opened 2018-02-26
Fix bug on PMGDQueryHandler which locks the graph

luisremis:
PMGDQueryHandler hangs when there is a fail command. This fixed the problem, but @vishakha041 please check if this is the right way to handle it.

### DIFF
--- a/src/PMGDQueryHandler.cc
+++ b/src/PMGDQueryHandler.cc
@@ -48,7 +48,7 @@ void PMGDQueryHandler::init()
 {
     std::string dbname = VDMSConfig::instance()
                         ->get_string_value("pmgd_path", "default_pmgd");
- 
+
     // Create a db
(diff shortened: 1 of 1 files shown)


src/PMGDQueryHandler.cc
```diff
@@ -169,6 +169,12 @@ int PMGDQueryHandler::process_query(const PMGDCmd *cmd,
         throw e;
     }
 
+    if (retval < 0) {
```
vishakha041: This is not needed here. It can go in the error_cleanup function since we already call it on error return.

luisremis: Won't work. error_cleanup can be called after unlocking _dblock, std::mutex has undefined behavior if you unlock a lock twice. 
The safe place to unlock is the same function we lock ( process_query(2) ).

vishakha041:
I have a separate fix based on Philip's input too. Will create a new pull request.

PR #9, opened 2018-03-21
Update README and Install/Run instructions

luisremis:


crstrong:
There still isn't mention of the need to create indexes first ... we really need to mention that somewhere.

Also, we should probably point out that you should only use run_server.sh when you haven't run the server before (ie, the database doesn't already exist) Using it otherwise is going to delete your existing database.

luisremis:
The indexes will be on some new wiki page on evaluation steps.

PR #14, opened 2018-04-09
Bug fixes

vishakha041:
Minor fixes to be consistent with API expectations.

### DIFF
--- a/src/RSCommand.cc
+++ b/src/RSCommand.cc
@@ -155,8 +155,7 @@ int AddEntity::construct_protobuf(PMGDQuery& query,
 {
     const Json::Value& cmd = jsoncmd[_cmd_name];
 
-    int node_ref = get_value<int>(cmd, "_ref",
-                                  query.get_available_reference());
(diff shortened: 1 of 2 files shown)


src/RSCommand.cc
```diff
@@ -155,8 +155,7 @@ int AddEntity::construct_protobuf(PMGDQuery& query,
 {
     const Json::Value& cmd = jsoncmd[_cmd_name];
 
-    int node_ref = get_value<int>(cmd, "_ref",
-                                  query.get_available_reference());
+    int node_ref = get_value<int>(cmd, "_ref", -1);
```
luisremis: this is not right. if link is present (if true on line 167), the node_ref will be -1 if the user did not set _ref. 
In other words, if the user does not use _ref and does use link, it will fail. 

One possible way to make it right would be to move the link if statement (line 167) up, and based on that set the node_ref.

vishakha041: Good point. I will address that and push new version.

PR #20, opened 2018-04-23
Feature: blobs on Entities

luisremis:
Add functionality to include a blob with an entity. 
Also, add flag on results block to indicate whether a blob (image, entity blob, etc) will be returned or not, assuming true as a default.

The commits will be squashed together when merging.


tests/python/TestImages.py
```diff
@@ -0,0 +1,357 @@
...
+
+        response, img_array = db.query(all_queries)
+        # print vdms.aux_print_json(response)
+
+        response = json.loads(response)
+        self.assertEqual(response[0]["FindImage"]["status"], 0)
+        self.assertEqual(response[1]["FindImage"]["status"], 0)
+        self.assertEqual(len(img_array), 0)
+
+    # This test is failing.
```
crstrong: Is this test still failing? We might want to take it out for now, until we figure out what's going on

src/RSCommand.cc
```diff
@@ -145,6 +151,14 @@ void RSCommand::add_link(PMGDQuery& query, const Json::Value& link,
 
 AddEntity::AddEntity() : RSCommand("AddEntity")
 {
+    _storage_blob = VDMSConfig::instance()
```
vishakha041: Shouldn't this variable be called _blob_path? This isn't a blob by itself.

src/RSCommand.cc
```diff
@@ -158,10 +172,29 @@ int AddEntity::construct_protobuf(PMGDQuery& query,
     int node_ref = get_value<int>(cmd, "_ref",
                                   query.get_available_reference());
 
+    // Modifiyng the existing properties that the user gives
+    // is a good option to make the AddNode more simple.
+    // This is not ideal since we are manupulating with user's
+    // input, but for now it is an acceptable solution.
+    Json::Value props = get_value<Json::Value>(cmd, "properties");
```
vishakha041: Is there a chance the user might add an entity without any properties? In which case you have to create a new block here?

src/RSCommand.cc
```diff
@@ -158,10 +172,29 @@ int AddEntity::construct_protobuf(PMGDQuery& query,
...
+
+    if (cmd.isMember("blob")) {
+        std::ostringstream oss;
+        oss << std::hex << VCL::get_int64();
+        std::string file_name = _storage_blob + oss.str();
+
+        props[VDMS_BLOB_PATH_PROP] = file_name;
+
+        std::ofstream file;
+        file.open(file_name);
```
vishakha041: Can this throw?

tests/python/TestEntitiesBlobs.py
```diff
@@ -0,0 +1,164 @@
...
+        response, res_arr = db.query(all_queries, [blob_arr])
+        response = json.loads(response)
+
+        self.assertEqual(response[0]["AddEntity"]["status"], 0)
+
+        constraints = {}
+        constraints["id"] = ["==", 45334]
+
+        results = {}
+        results["blob"] = True
```
vishakha041: I haven't yet seen a commit supporting blobs in Find. How come this test works?

src/RSCommand.cc
```diff
@@ -246,12 +246,18 @@ int FindEntity::construct_protobuf(
 {
     const Json::Value& cmd = jsoncmd[_cmd_name];
 
+    Json::Value results = get_value<Json::Value>(cmd, "results");
```
vishakha041: Does this work even when no results block is given with FindEntity?

src/ImageCommand.cc
```diff
@@ -164,7 +164,10 @@ int FindImage::construct_protobuf(
     const Json::Value& cmd = jsoncmd[_cmd_name];
 
     Json::Value results = get_value<Json::Value>(cmd, "results");
-    results["list"].append(VDMS_IM_PATH_PROP);
+
+    if (get_value<bool>(results, "return_blob", true)){
```
vishakha041: This is called just blob for FindEntity. I like the shorter one but just keep them consistent.

vishakha041: Never mind. I think you made them consistent later though FindEntity now has a default of true with return_blob. Two concerns:
- we don't say return_avg or return_sum. So blob might be more suitable when comparing with the other things. Results already implies returning stuff.
- For images, it is natural to expect blobs but not so for entities. It should have a default of false.

tests/python/TestImages.py
vishakha041: Second that

src/ImageCommand.cc
```diff
@@ -36,9 +36,9 @@
 
 using namespace VDMS;
 
-#define VDMS_IM_TAG           "AT:IMAGE"
-#define VDMS_IM_PATH_PROP     "imgPath"
-#define VDMS_IM_EDGE          "AT:IMG_LINK"
+#define VDMS_IM_TAG           "VDMS:IMG"
```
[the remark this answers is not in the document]
luisremis: I wanted to avoid making a pull request with one line of code, that's why I added it here.

src/RSCommand.cc
luisremis: this is following the same convention on images. where the path to pngs/jpg etc is _storage_format

src/RSCommand.cc
luisremis: no need. if props is empty, nothing will be done with it.

src/RSCommand.cc
luisremis: yes, same as the properties above. it will return an empty object, we check for that in PMGDQuery

src/ImageCommand.cc
luisremis: This is what I suggested, but the agreement was that it was either default for all, or for none, to make it consistent, and since it made sense for images to be the default, we decided to also make it the default for entities as well. 

Also, it make sense to make true as the default for all, as if you add an entity with a blob, you are more likely to be interested in getting the blob when findEntity. 

I will leave it as default=true for all, unless we decide otherwise then.

luisremis: also, I am changing the name to "blob" in all cases.

src/ImageCommand.cc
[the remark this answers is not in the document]
luisremis: this is a separate commit. Is separate commit not enough? Do we really need to make a separate pull requests for this?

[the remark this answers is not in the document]
luisremis: This is not a bug fix, so I don't see it is ok to put it in a bug fix branch but not ok to put it on a feature branch. 

Will create a new pull request with this change after blob feature is merged for the sake of our mental health.

luisremis:
Creating a new pull request.

PR #21, opened 2018-04-24
Add Descriptors and Classification

luisremis:
This add the AddDescriptorSet, AddDescriptor, and ClassifyDescriptor calls. 
I will split this commits, there are a couple a cases where it makes sense to separate some things that are bug-fixing. 

This is based on feature_blobAddentity so it won't show the changes of that branch (that is another on-going pull request)


luisremis:
this needs some more work. Will create a new PR later.

PR #29, opened 2018-05-12
Api extension

vishakha041:
Added some long pending extensions to our API. The wiki entries are at the following links.

https://github.com/IntelLabs/vdms/wiki/UpdateEntity
https://github.com/IntelLabs/vdms/wiki/UpdateImage
https://github.com/IntelLabs/vdms/wiki/FindConnection

### DIFF
--- a/src/PMGDIterators.cc
+++ b/src/PMGDIterators.cc
@@ -33,30 +33,24 @@
 
 using namespace VDMS;
 
-bool PMGDQueryHandler::ReusableNodeIterator::_next()
+namespace VDMS {
+
+template <>
+PMGDQueryHandler::ReusableIterator<PMGD::Edge, PMGD::EdgeIterator>::
+ReusableIterator() :
+        _ti(NULL),
+        _it(_traversed.end())
 {
-    if (_it != _traversed.end()) {
-        ++_it;
-        if (_it != _traversed.end())
-            return true;
-    }
-    if (bool(_ni)) {
-        _it = _traversed.insert(_traversed.end(), &*_ni);
-        _ni.next();
-        return true;
-    }
-    return false;
 }
 
-// TODO It might be possible to avoid this if the first iterator
-// was build out of an index sorted on the same key been sought here.
-// Hopefully that won't happen.
-void PMGDQueryHandler::ReusableNodeIterator::sort(PMGD::StringID sortkey)
+template<>
+void PMGDQueryHandler::ReusableIterator<PMGD::Edge, PMGD::EdgeIterator>::add(PMGD::Edge *e)
 {
-    // First finish traversal
-    traverse_all();
-    _traversed.sort(compare_propkey{sortkey});
-    _it = _traversed.begin();
+    // Easiest to add to the end of list. If we are in middle of
+    // traversal, then this edge might get skipped. Use this function
+    // with that understanding ***
+    _traversed.insert(_traversed.end(), e);
+}
 }
 
 bool PMGDQueryHandler::MultiNeighborIteratorImpl::_next()
@@ -87,32 +81,6 @@ bool PMGDQueryHandler::MultiNeighborIteratorImpl::next()
             return true;
     }
     return _next();
-}
-
-bool PMGDQueryHandler::ReusableEdgeIterator::_next()
-{
-    if (_it != _traversed.end()) {
-        ++_it;
-        if (_it != _traversed.end())
-            return true;
-    }
-    if (bool(_ei)) {
-        _it = _traversed.insert(_traversed.end(), &static_cast<PMGD::Edge &>(*_ei));
-        _ei.next();
-        return true;
-    }
-    return false;
-}
-
-// TODO It might be possible to avoid this if the first iterator
-// was build out of an index sorted on the same key been sought here.
-// Hopefully that won't happen.
-void PMGDQueryHandler::ReusableEdgeIterator::sort(PMGD::StringID sortkey)
-{
-    // First finish traversal
-    traverse_all();
-    _traversed.sort(compare_propkey{sortkey});
-    _it = _traversed.begin();
 }
 
 bool PMGDQueryHandler::NodeEdgeIteratorImpl::next()
@@ -155,8 +123,6 @@ bool PMGDQueryHandler::NodeEdgeIteratorImpl::check_predicates()
     }
     if (_check_dest &&
             _dest_nodes.find(&(e->get_destination()) ) == _dest_nodes.end())
-        return false;;
+        return false;
     return true;
 }
-
-
--- a/src/PMGDIterators.h
+++ b/src/PMGDIterators.h
@@ -38,21 +38,37 @@
 #include "SearchExpression.h"
 
 namespace VDMS {
-    class PMGDQueryHandler::ReusableNodeIterator
+
+    template <typename T, typename Ti>
+    class PMGDQueryHandler::ReusableIterator
     {
         // Iterator for the starting nodes.
-        PMGD::NodeIterator _ni;
+        Ti _ti; // Type Iterator
 
-        // TODO Is list the best data structure if we could potentially
-        // sort?
-        std::list<PMGD::Node *> _traversed;
+        // TODO Is list the best data structure
+        // if we could potentially sort?
+        typedef std::list<T* > base_container;
+        base_container _traversed;
 
         // Current postion of list iterator
-        std::list<PMGD::Node *>::iterator _it;
+        typedef typename base_container::iterator list_iterator;
+        list_iterator _it;
 
-        bool _next();
+        bool _next() {
+            if (_it != _traversed.end()) {
+                ++_it;
+                if (_it != _traversed.end())
+                    return true;
+            }
+            if (bool(_ti)) {
+                _it = _traversed.insert(_traversed.end(), &static_cast<T &>(*_ti));
+                _ti.next();
+                return true;
+            }
+            return false;
+        }
 
-        PMGD::Node *ref()
+        T *ref()
         {
             if (!bool(*this))
                 throw PMGDException(NullIterator, "Null impl");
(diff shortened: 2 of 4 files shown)


vishakha041:
https://github.com/IntelLabs/vdms/wiki/UpdateConnection

vishakha041:
This branch is rebased on top of the fixes branch. Please do review the new edge iterators carefully. I ended up adding all the functionalities for edges except conditional add. I am not sure there is a need for it and not sure enough time.

src/PMGDQueryHandler.cc
```diff
@@ -241,6 +244,50 @@ int PMGDQueryHandler::add_node(const protobufs::AddNode &cn,
...
+    long id = un.identifier();
+    bool query = un.has_query_node();
+
+    // Make sure there is either an id or a query node
+    if (id < 0 && !query) {
+        set_response(response, PMGDCmdResponse::Error, "No way to find update node\n");
+        return -1;
+    }
+
+    auto it = _cached_nodes.find(id);
```
philiplantz: Shouldn't this check whether id >= 0 before calling find? (See suggestion below.)

src/PMGDQueryHandler.cc
```diff
@@ -241,6 +244,50 @@ int PMGDQueryHandler::add_node(const protobufs::AddNode &cn,
...
 }
 
+int PMGDQueryHandler::update_node(const protobufs::UpdateNode &un,
+                    protobufs::CommandResponse *response)
+{
+    long id = un.identifier();
+    bool query = un.has_query_node();
+
+    // Make sure there is either an id or a query node
+    if (id < 0 && !query) {
```
philiplantz: Is it also an error if /both/ are set?

src/PMGDQueryHandler.cc
```diff
@@ -241,6 +244,50 @@ int PMGDQueryHandler::add_node(const protobufs::AddNode &cn,
...
+            set_response(response, PMGDCmdResponse::Error, "Undefined _ref value used in update\n");
+            return -1;
+        }
+        else {
+            query_node(un.query_node(), response);
+            if (response->error_code() != PMGDCmdResponse::Success)
+                return -1;
+            it = _cached_nodes.find(un.query_node().identifier());  // id could have been -ve
+        }
+    }
```
philiplantz: Perhaps lines 251-271 should be something like this?

```
if (id >= 0) {  
    it = _cached_nodes.find(id);  
    if (it == cached_nodes.end()) {
        set_response(response, PMGDCmdResponse::Error, "Undefined _ref value used in update\n");
        return -1;
    }
}
else if (un.has_query_node()) {
    query_node(un.query_node(), response);
    if (response->error_code()  != PMGDCmdResponse::Success)
        return -1;
    it = _cached_nodes.find(un.query_node().identifier());
}
else {
    set_response(response, PMGDCmdResponse::Error, "No way to find update node\n");
    return -1;
}
```

src/PMGDQueryHandler.cc
```diff
@@ -241,6 +244,50 @@ int PMGDQueryHandler::add_node(const protobufs::AddNode &cn,
...
+    if (it == _cached_nodes.end()) {
+        if (!query) {
+            set_response(response, PMGDCmdResponse::Error, "Undefined _ref value used in update\n");
+            return -1;
+        }
+        else {
+            query_node(un.query_node(), response);
+            if (response->error_code() != PMGDCmdResponse::Success)
+                return -1;
+            it = _cached_nodes.find(un.query_node().identifier());  // id could have been -ve
```
philiplantz: Is "-ve" supposed to mean "negative"?

src/PMGDIterators.h
```diff
@@ -132,4 +134,152 @@ namespace VDMS {
...
+
+        bool next();
+        PMGD::EdgeRef *ref() { return &(**_edge_it); }
+        PMGD::StringID get_tag() const { return (*_edge_it)->get_tag(); }
+        PMGD::Node &get_source() const { return (*_edge_it)->get_source(); }
+        PMGD::Node &get_destination() const { return (*_edge_it)->get_destination(); }
+        PMGD::Edge *get_edge() const { return &static_cast<PMGD::Edge &>(**_edge_it); }
+    };
+
+    class PMGDQueryHandler::ReusableEdgeIterator
```
philiplantz: This is almost an exact duplicate of ReusableNodeIterator. Seems like a good place for a template?

src/PMGDIterators.h
```diff
@@ -132,4 +134,152 @@ namespace VDMS {
 
         PMGD::Node *ref() { return &**_neighb_i; }
     };
+
+    class PMGDQueryHandler::NodeEdgeIteratorImpl : public PMGD::EdgeIteratorImplIntf
```
philiplantz: I need to understand the purpose of this and how it is used in order to review it.

src/PMGDQueryHandler.cc
```diff
@@ -365,6 +368,51 @@ int PMGDQueryHandler::add_edge(const protobufs::AddEdge &ce,
...
+    long updated = 0;
+    for ( ; *eit; eit->next()) {
+        Edge &e = **eit;
+        updated++;
+        for (int i = 0; i < ue.properties_size(); ++i) {
+            const protobufs::Property &p = ue.properties(i);
+            set_property(e, p);
+        }
+        for (int i = 0; i < ue.remove_props_size(); ++i)
+            e.remove_property(ue.remove_props(i).c_str());
```
philiplantz: If many nodes/edges are being updated, it would be advantageous to get the StringIDs for the properties in advance instead of converting each property name to a StringID every time it is used.

src/PMGDQuery.cc
```diff
@@ -463,6 +463,8 @@ void PMGDQuery::UpdateNode(int ref,
                         const Json::Value& constraints,
                         bool unique)
 {
+    _readonly = false;
```
philiplantz: This bug fix for UpdateNode should be merged into the commit where UpdateNode was added.

src/PMGDIterators.h
```diff
@@ -132,4 +134,152 @@ namespace VDMS {
...
+            if (_src_ni == NULL) {
+                if (_dest_ni == NULL)
+                    _pred_start = 1;
+                else {
+                    _dir = PMGD::Direction::Incoming;
+                    _src_ni = _dest_ni;
+                    _dest_ni = NULL;
+                }
+            }
+            return (_src_ni == NULL || !bool(*_src_ni)) ? _expr.db().get_edges(_expr.tag(),
```
luisremis: split this into if/elses.

luisremis: even if this works, !bool(*_src_ni) will never be empty because of how the code is right now, but we should change in the future because we want to continue with the transaction even if some querynode did not find anything. 

return null if the src iterators is empty.

src/PMGDIterators.h
```diff
@@ -132,4 +134,152 @@ namespace VDMS {
...
+        NodeEdgeIteratorImpl(const SearchExpression &expr,
+                             ReusableNodeIterator *src_ni,
+                             ReusableNodeIterator *dest_ni = NULL)
+            : _expr(expr), _num_predicates(_expr.num_predicates()),
+              _src_ni(src_ni), _dest_ni(dest_ni),
+              _pred_start(0), _check_dest(false),
+              _edge_it(new PMGD::EdgeIterator(return_iterator()))
+        {
+            if (_dest_ni != NULL) {
+                for (; bool(*_dest_ni); _dest_ni->next())
```
luisremis: maybe you can use the insert call here that allows you to specify begin and end, and avoid the for loop.

src/PMGDIterators.cc
```diff
@@ -89,3 +89,74 @@ bool PMGDQueryHandler::MultiNeighborIteratorImpl::next()
...
+{
+    PMGD::Edge *e = get_edge();
+    for (std::size_t i = _pred_start; i < _num_predicates; i++) {
+        PMGD::PropertyFilter<PMGD::Edge> pf(_expr.predicate(i));
+        if (pf(*e) == PMGD::DontPass)
+            return false;
+    }
+    if (_check_dest &&
+            _dest_nodes.find(&(e->get_destination()) ) == _dest_nodes.end())
+        return false;;
```
luisremis: double semicolon, even if cool, not necessary.

src/PMGDIterators.h
```diff
@@ -132,4 +134,152 @@ namespace VDMS {
...
+        struct compare_propkey
+        {
+            PMGD::StringID _propid;
+            bool operator()(const PMGD::Edge *e1, const PMGD::Edge *e2)
+              { return e1->get_property(_propid) < e2->get_property(_propid); }
+        };
+
+    public:
+        // Make sure this is not auto-declared. The move one won't be.
+        ReusableEdgeIterator(const ReusableEdgeIterator &) = delete;
```
luisremis: we can make the reusable iterator a template and have the same for edge/nodes.

tests/json_queries.cc
```diff
@@ -100,6 +100,64 @@ TEST(AddImage, simpleAdd)
 
     EXPECT_EQ(json_response[0]["AddImage"]["status"].asString(), "0");
     VDMSConfig::destroy();
+    PMGDQueryHandler::destroy();
+}
+
+TEST(UpdateEntity, simpleAddUpdate)
```
luisremis: I am inclined to move all this testing to python. 
Having all this tests in python make great tutorials. 
Since we are testing at user-level query (means black-box test), I think it make more sense to just do those in python, and leave the unit testing in C++ (the PMGDQueryHandler testing is a perfect example of unit testing)
We should no longer keep adding stuff to json_queries.cc (which reads the query from a file) and migrate those tests to Python (most of them are already in Python in any case)

src/PMGDQuery.cc
```diff
@@ -441,16 +444,58 @@ void PMGDQuery::AddNode(int ref,
...
+    }
+
+    for (auto it = remove_props.begin(); it != remove_props.end(); it++) {
+        std::string *r_key= un->add_remove_props();
+        *r_key = (*it).asString();
+    }
+
+    if(!constraints.isNull()) {
+        PMGDQueryNode *qn = un->mutable_query_node();
+        qn->set_identifier(ref < 0 ? get_available_reference() : ref);
```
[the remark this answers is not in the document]
luisremis: Not sure why this is, but it won't work if I remove it. Look at it later.

src/RSCommand.cc
```diff
@@ -205,13 +205,57 @@ int AddEntity::construct_protobuf(PMGDQuery& query,
...
+    int grp_id,
+    Json::Value& error)
+{
+    const Json::Value& cmd = jsoncmd[_cmd_name];
+
+    int node_ref = get_value<int>(cmd, "_ref", -1);
+
+    query.UpdateNode(
+            node_ref,
+            get_value<std::string>(cmd, "class"),
```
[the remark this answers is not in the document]
luisremis: this line is actually wrong, because it will default to empty string for the class, fixed with cmd["class"].asString()

utils/src/api_schema/api_schema.json
```diff
@@ -247,6 +316,18 @@
       "additionalProperties": false
     },
 
+    "UpdateImage": {
+      "properties": {
+        "_ref":         { "$ref": "#/definitions/refInt" },
+        //"format":       { "$ref": "#/definitions/formatString" },
```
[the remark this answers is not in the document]
luisremis: nah, we can leave this for the future.

utils/src/api_schema/api_schema.json
```diff
@@ -223,6 +266,19 @@
       "additionalProperties": false
     },
 
+    "UpdateConnection": {
+      "properties": {
+        "class":        { "type": "string" },
```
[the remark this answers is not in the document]
luisremis: It seems class is not required, so no need to change this. I will update the wiki to reflect that class is optional.

tests/json_queries.cc
luisremis: We will leave this for now, and keep adding Python tests as well.

src/PMGDQueryHandler.cc
luisremis: Fixed

src/PMGDQueryHandler.cc
luisremis: Fixed.

src/PMGDQueryHandler.cc
luisremis: Fixed

src/PMGDQueryHandler.cc
luisremis: True. Added a comment on this on the code.

PR #30, opened 2018-05-21
Fixes

vishakha041:
In case of errors, the lack of PMGDQueryHandler pointer cleanup was causing problems. And not being able to set the number of attempts for RWLock was annoying for some application use cases. This pull request address both the problems.

### DIFF
--- a/src/PMGDQueryHandler.cc
+++ b/src/PMGDQueryHandler.cc
@@ -386,6 +386,8 @@ int PMGDQueryHandler::query_node(const protobufs::QueryNode &qn,
     if (!bool(ni)) {
         set_response(response, PMGDCmdResponse::Empty,
                        "Null search iterator\n");
+        if (has_link)
+            start_ni->reset();
         return -1;
     }
 
@@ -413,6 +415,8 @@ int PMGDQueryHandler::query_node(const protobufs::QueryNode &qn,
         if (bool(*tni)) {  // Not unique and that is an error here.
             set_response(response, PMGDCmdResponse::NotUnique,
                            "Query response not unique\n");
+            if (has_link)
+                start_ni->reset();
             delete tni;
             return -1;
         }
--- a/src/PMGDQueryHandler.cc
+++ b/src/PMGDQueryHandler.cc
@@ -49,12 +49,14 @@ void PMGDQueryHandler::init()
 {
     std::string dbname = VDMSConfig::instance()
                         ->get_string_value("pmgd_path", "default_pmgd");
+    unsigned attempts = VDMSConfig::instance()
+                        ->get_int_value("max_lock_attempts", RWLock::MAX_ATTEMPTS);
  
     // Create a db
     _db = new PMGD::Graph(dbname.c_str(), PMGD::Graph::Create);
 
     // Create the query handler here assuming database is valid now.
-    _dblock = new RWLock();
+    _dblock = new RWLock(attempts);
 }
 
 void PMGDQueryHandler::destroy()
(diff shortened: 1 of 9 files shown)


src/RWLock.h
```diff
@@ -52,15 +52,15 @@ namespace VDMS {
...
-        static const unsigned MAX_ATTEMPTS      = 10;
+        static const size_t MAX_BACKOFF_DELAY = 500000000;
 
         uint16_t xadd(volatile uint16_t &m, uint16_t v)
             { return ::xadd<uint16_t>(m, v); }
         void atomic_and(volatile uint16_t &m, uint16_t v)
             { ::atomic_and<uint16_t>(m, v); }
 
         volatile uint16_t _rw_lock;
+        unsigned _max_attempts;
```
philiplantz: This should be const (since it is initialized in the constructor). This may allow compiler to improve code generation in the methods where it is used.

philiplantz: I don't see where _max_attempts is used.

src/PMGDQueryHandler.h
```diff
@@ -119,6 +119,7 @@ namespace VDMS {
 
     public:
         static void init();
+        static void destroy();
```
philiplantz: Why isn't this a destructor?

src/PMGDQueryHandler.cc
```diff
@@ -57,6 +57,16 @@ void PMGDQueryHandler::init()
     _dblock = new RWLock();
 }
 
+void PMGDQueryHandler::destroy()
+{
+    if (_db) {
```
philiplantz: This test isn't needed.

src/RWLock.h
philiplantz: I guess it's not clear that this is the comment that caused me to withhold approval. The other comments are questions/suggestions.

src/PMGDQueryHandler.h
vishakha041: just pairs up with an init

philiplantz: Okay, I can see that. But init is needed (I assume) because the information required to perform initialization isn't available at construction time. The same concern doesn't apply to using a destructor, and I think performing this cleanup work in the destructor would be preferable.

philiplantz: As you pointed out, the object is a static member, so the destructor isn't called at the time when this destroy function is called.

src/RWLock.h
```diff
@@ -80,7 +80,9 @@ namespace VDMS {
         }
 
     public:
-        RWLock() : _rw_lock(0) {}
+        static const unsigned MAX_ATTEMPTS      = 10;
```
philiplantz: This should be called DEFAULT_MAX_ATTEMPTS.

tests/python/TestEntities.py
```diff
@@ -98,13 +98,13 @@ def findEntity(self, thID):
...
 
     def test_runMultipleAdds(self):
-
         simultaneous = 1000;
         thread_arr = []
         for i in range(1,simultaneous):
             thread_add = Thread(target=self.addEntity,args=(i,) )
             thread_add.start()
             thread_arr.append(thread_add)
+            time.sleep(0.002)
```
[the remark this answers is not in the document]
vishakha041: I think we can create a separate issue. What I noticed here is....msync increases how long PMGD takes. That means the reader writer lock has more timeouts and so more threads wait. That somehow seems to be messing with some queue that starts denying newer threads from even getting connected. Needs more exploring which I don't want to hold up this pull request for.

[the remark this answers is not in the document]
vishakha041: Also, one of the things we need to figure now at the VDMS layer is that timeout based locking is bound to cause more failures since threads don't get to wait however long. Do we introduce internal retries or let the client retry? We should of course try to figure out the right values as we run more experiments but just saying.

PR #34, opened 2018-08-14
Feature descriptors

luisremis:
Adds Descriptors functionality:
AddDescriptorSet
AddDescriptor
FindDescriptor
ClassifyDescriptor

### DIFF
--- a/src/DescriptorsCommand.cc
+++ b/src/DescriptorsCommand.cc
@@ -1,0 +1,887 @@
+/**
+ * @file   DescriptorsCommand.cc
+ *
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2017 Intel Corporation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction,
+ * including without limitation the rights to use, copy, modify,
+ * merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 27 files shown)


src/DescriptorsCommand.cc
```diff
@@ -0,0 +1,902 @@
...
+        findDesc.removeMember("entities");
+
+        convert_properties(entities);
+
+        std::cout << json_responses << std::endl;
+        std::cout << (*ids).size() << std::endl;
+
+        // This should never happen as ids will be have all (and only)
+        // the valid elements at this point
+        assert(entities.size() == (*ids).size());
```
crstrong: I know we were having an issue with this assert when all the features were identical within the vector. I know that they shouldn't all be identical, but is assert the right response? (Maybe we should throw here instead?)

src/DescriptorsManager.h
```diff
@@ -0,0 +1,64 @@
...
+        static DescriptorsManager* instance();
+
+        /**
+         *  Handles descriptors and lock for the descriptor
+         *  This is a blocking call until the descriptor is free
+         *
+         *  @param path  Path to the descriptor set
+         */
+        VCL::DescriptorSet* get_descriptors_handler(std::string path);
+        void flush();
```
crstrong: When does this get called?

src/DescriptorsCommand.cc
```diff
@@ -0,0 +1,902 @@
...
+        eng = VCL::TileDBDense;
+    else if (eng_str == "TileDBSparse")
+        eng = VCL::TileDBSparse;
+    else
+        throw ExceptionCommand(DescriptorSetError, "Engine not supported");
+
+    // This can be a problem. If this fails, we will have a node in pmgd
+    // linking to this broken directory. We can probably set up a mechanism
+    // to fix a broken link when detected later, same with images.
+    // For now, we use the defaul faiss index.
```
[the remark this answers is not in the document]
luisremis: Those are disconnected statements, yes. I will put a line in between to avoid confusion.

src/DescriptorsManager.h
luisremis: This get's called in the destructor of the server.

src/DescriptorsCommand.cc
luisremis: This assert does not control that, it controls that pmgd response is consistent with the information of the search. But since it is redundant because of previous controls, we can get rid of it.

PR #41, opened 2018-10-15
Add flip/rotate

luisremis:
Adding very simple operations that some users requested for VDMS.
I hope this PR can be used as reference for contributors on how to add new operations.

### DIFF
--- a/src/ImageCommand.cc
+++ b/src/ImageCommand.cc
@@ -63,6 +63,13 @@ void ImageCommand::enqueue_operations(VCL::Image& img, const Json::Value& ops)
                         get_value<int>(op, "width"),
                         get_value<int>(op, "height") ));
         }
+        else if (type == "flip") {
+            img.flip(get_value<int>(op, "code"));
+        }
+        else if (type == "rotate") {
+            img.rotate(get_value<double>(op, "angle"),
+                       get_value<bool>(op, "resize"));
+        }
         else {
             throw ExceptionCommand(ImageError, "Operation not defined");
         }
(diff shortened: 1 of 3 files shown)


src/ImageCommand.cc
```diff
@@ -63,6 +63,13 @@ void ImageCommand::enqueue_operations(VCL::Image& img, const Json::Value& ops)
                         get_value<int>(op, "width"),
                         get_value<int>(op, "height") ));
         }
+        else if (type == "flip") {
```
vishakha041: This is getting long enough to make a switch case out of this but that will need some JSON level integer percolation instead of string. Anyway, something to think about as an optimization for us going forward as more operations get added. That's why don't make this a standard pull request way for adding ops. We need to think more.

client/python/vdms/__init__.py
```diff
@@ -0,0 +1,4 @@
+name = "vdms"
```
vishakha041: For some reason I am seeing the python pull request merged with flip feature PR. Could you make sure they are separately merged on develop?

utils/src/api_schema/api_schema.json
```diff
@@ -173,6 +175,28 @@
...
+      "type": "object",
+      "properties": {
+        "type":   { "enum": [ "flip" ] },
+        "code":   { "type": "integer" }
+      },
+      "required": ["type", "code"],
+      "additionalProperties": false
+    },
+
+    "operationRotate": {
```
vishakha041: Could you please make sure the wiki page is updated to reflect this. I haven't checked if it is

client/python/vdms/__init__.py
luisremis: this will go away, yes. it is here only because the python PR is also pending.

src/ImageCommand.cc
luisremis: good idea, we can have a string table here.

vishakha041:
I think it starts becoming beneficial after a certain number of elements. For now this might be ok. Best would be to do some of these conversions at client library end

Issue #52, opened 2018-11-26
Sort by ascending/descending


luisremis:
External request: Are we able to specify the direction of the sort, e.g., ascending vs descending? As by default is sorts in ascending order and we want to sort by last modified date in descending order.

luisremis:
Solved in #53

PR #53, opened 2018-11-26
Feature sort descending

luisremis:
Solve #52


utils/src/api_schema/api_schema.json
```diff
@@ -113,13 +113,14 @@
...
-        "limit":    { "$ref": "#/definitions/positiveInt" },
-        "sort":     { "type": "string" },
-        "blob":     { "type": "boolean" }
+        "list":       { "type": "array" },
+        "average":    { "type": "string" },
+        "count":      { "type": "string" },
+        "sum":        { "type": "string" },
+        "limit":      { "$ref": "#/definitions/positiveInt" },
+        "sort":       { "type": "string" },
+        "descending": { "type": "boolean" },
```
vishakha041: Is there no way to make this part of the sort command? Descend keyword by itself doesn't make sense. It is useful only in the sort context.

tests/python/TestEntities.py
```diff
@@ -205,3 +205,78 @@ def test_addEntityWithLink(self):
 
         self.assertEqual(response[0]["AddEntity"]["status"], 0)
         self.assertEqual(response[1]["AddEntity"]["status"], 0)
+
+    def test_FindAscending(self):
```
vishakha041: Shouldn't this be called FindDescending? Or FindWithSortOrder :)

utils/src/api_schema/api_schema.json
luisremis: yep, I was divided about this as well. Adding a "sort block" that has a prop for sorting and the option of using order = "ascending" or order = "descending" was another option. We an go with that, I think I like it more, but it was 50/50.

vishakha041: Will it still be possible to have a default so you can only specify the key ?

luisremis: Ok. 
The Options are:

1)  (breaks backward compatibility)
"sort": { 
    "key": "property_used_for_sorting", 
    "order": "ascending" // or "descending", with the default to one of them. 
}

2) As it is implemented, just adds the order as an extra element of the results block, that will only be used if "sort" is also specified. 

3) We can change "descending" = [True, False] to "order" = ["ascending", "descending"]. I like this option more, which does not break compatibility, it just adds more flexibility.

philiplantz: Is there no way to support either a value or a block, as:
    "sort" : "\<key\>"
and
    "sort" : { "key" : "\<key\>", "order" : "descending" }

If not, I'm inclined to go with the backward compatible option, with the addition
     "sort-order" = ["ascending","descending"]

luisremis:
I like what you propose @philiplantz. For some reason I have a memory that we cannot do that. 
Let me try it harder.

vishakha041:
I like the option Philip has mentioned. Otherwise option 1 from Luis. It will break backward compatibility but will open the way for allowing sort on multiple keys and so on in the near future.

luisremis:
Sorting multiple keys is independent for of the having a sort block or not. I will go with option 2 to avoid braking compatibility for now, and change the keywords following Philip suggestion which is much nicer.

mwahle:
When can we have this in the master branch? Or even released?

PR #74, opened 2019-01-28
Basic Video Support

luisremis:
Add basic support for video. Updates Image formats definitions.


src/VideoCommand.cc
```diff
@@ -0,0 +1,352 @@
...
+        else if (type == "resize") {
+             video.resize(get_value<int>(op, "height"),
+                           get_value<int>(op, "width") );
+
+        }
+        else if (type == "crop") {
+            video.crop(VCL::Rectangle (
+                        get_value<int>(op, "x"),
+                        get_value<int>(op, "y"),
+                        get_value<int>(op, "width"),
```
vishakha041: Inconsistent with the wiki page. Also, the wiki page has some JSON formatting issues in example. For a crop, number frames is missing and it is called w and h which is cool. I am not clear on the x,y,z naming.

src/VideoCommand.cc
```diff
@@ -0,0 +1,352 @@
...
+            video.interval(
+                VCL::Video::FRAMES,
+                get_value<int>(op, "start"),
+                get_value<int>(op, "stop"),
+                get_value<int>(op, "step"));
+
+        }
+        else if (type == "resize") {
+             video.resize(get_value<int>(op, "height"),
+                           get_value<int>(op, "width") );
```
vishakha041: w and h on wiki. Also, I think the operations could be explained a little more in the context of video on the wiki. + an example of how a blob could be specified.

src/VideoCommand.cc
```diff
@@ -0,0 +1,352 @@
...
+        }
+        else {
+            throw ExceptionCommand(ImageError, "Operation not defined");
+        }
+    }
+}
+
+VCL::Video::Codec VideoCommand::string_to_codec(std::string& codec)
+{
+    if (codec == "h263") {
```
vishakha041: WIki page should list supported codecs btw and the defaults

src/VideoCommand.cc
```diff
@@ -0,0 +1,352 @@
...
+
+    VCL::Video video((void*)blob.data(), blob.size());
+
+    if (cmd.isMember("operations")) {
+        enqueue_operations(video, cmd["operations"]);
+    }
+
+    // The container and codec are checked by the schema.
+    // We default to mp4 and h264, if not specified
+    std::string container  = get_value<std::string>(cmd, "container", "mp4");
```
vishakha041: Missing & here and other places?

src/VideoCommand.cc
```diff
@@ -0,0 +1,352 @@
...
+}
+
+//========= UpdateImage definitions =========
+
+UpdateVideo::UpdateVideo() : VideoCommand("UpdateVideo")
+{
+}
+
+int UpdateVideo::construct_protobuf(
+    PMGDQuery& query,
```
vishakha041: No update wiki page

vishakha041: So its not clear what update will support updating for videos.

src/VideoCommand.cc
```diff
@@ -0,0 +1,352 @@
...
+    Json::Value& responses,
+    const Json::Value& json,
+    protobufs::queryMessage &query_res,
+    const std::string &blob)
+{
+    assert(responses.size() == 1);
+
+    Json::Value ret;
+
+    // TODO In order to support "codec" or "operations", we could
```
vishakha041: Is this consistent with what images let you do for update?

src/defines.h
```diff
@@ -81,3 +81,9 @@
 #define VDMS_ROI_COORD_Y_PROP "VD:y1"
 #define VDMS_ROI_WIDTH_PROP   "VD:width"
 #define VDMS_ROI_HEIGHT_PROP  "VD:height"
+
+// Videos
+
+#define VDMS_VID_TAG           "VD:VID"
```
vishakha041: Inconsistent naming

utils/SConscript
```diff
@@ -60,14 +60,13 @@ chrono_cc = ['src/chrono/Chrono.cc']
...
 
 utils_env.ParseConfig('pkg-config --cflags --libs protobuf')
-utils_env.SharedLibrary('vdms-utils', utils_source_files)
+ulib = utils_env.SharedLibrary('vdms-utils', utils_source_files)
 
 # Comm Testing
 comm_test_env = Environment(CPPPATH=['include/comm'],
-                            CXXFLAGS="-std=c++11")
+                            CXXFLAGS="-std=c++11",
+                            LIBS = [ulib, 'pthread', 'gtest']
```
vishakha041: These again look like separate scons changes not part of video necessarily

src/VideoCommand.cc
```diff
@@ -0,0 +1,352 @@
...
+    const Json::Value& cmd = jsoncmd[_cmd_name];
+
+    Json::Value results = get_value<Json::Value>(cmd, "results");
+
+    // Unless otherwhise specified, we return the blob.
+    if (get_value<bool>(results, "blob", true)){
+        results["list"].append(VDMS_VID_PATH_PROP);
+    }
+
+    std::cout << "hello" << std::endl;
```
vishakha041: Remove this

src/VideoCommand.cc
```diff
@@ -0,0 +1,352 @@
...
+            if (!cmd.isMember("operations") &&
+                !cmd.isMember("container")  &&
+                !cmd.isMember("codec"))
+            {
+                // Return video as is.
+                std::ifstream ifile(video_path, std::ifstream::in);
+                ifile.seekg(0, std::ios::end);
+                size_t encoded_size = (long)ifile.tellg();
+                ifile.seekg(0, std::ios::beg);
+
```
vishakha041: Isn't this equivalent to asking VCL to read video? Seems like code repeat somehow

utils/src/api_schema/api_schema.json
```diff
@@ -452,8 +521,8 @@
       "properties": {
         "_ref":        { "$ref": "#/definitions/refInt" },
         "link":        { "$ref": "#/definitions/blockLink" },
-        "operations":  { "$ref": "#/definitions/blockOperations" },
-        "format":      { "$ref": "#/definitions/formatString" },
+        "operations":  { "$ref": "#/definitions/blockImageOperations" },
```
vishakha041: Aren't these changes part of the previous commit? It probably breaks some test in the previous commit.

tests/python/TestVideos.py
```diff
@@ -0,0 +1,335 @@
...
+
+        all_queries = []
+
+        constraints = {}
+        constraints["name"] = ["==", prefix_name + str(0)]
+
+        props = {}
+        props["name"] = "simg_update_0"
+
+        img_params = {}
```
vishakha041: Do you need me to review tests? I am hoping you have put checks and balances here to make sure videos get stored and rertieved correctly without losing out data :)

src/VDMSConfig.h
```diff
@@ -69,6 +70,7 @@ namespace VDMS{
         std::string path_jpg;
         std::string path_tdb;
         std::string path_blobs;
+        std::string path_videos;
```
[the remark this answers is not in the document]
luisremis: we will keep them consistent for this commit. it is consistent on what it is for the user and what it is internal, so I am ok with the names.

src/VideoCommand.cc
luisremis: Cool, glad you realized. Would you update the wiki? you have full access. 
I have just made sure that it is consistent with the JSON schema, so you can use the code as reference for updating the wiki.

src/VideoCommand.cc
luisremis: same as above.

src/VideoCommand.cc
luisremis: same as above.

src/VideoCommand.cc
luisremis: this is just to keep it the same as image. For now, it just updates the properties (same as images)

src/VideoCommand.cc
luisremis: yeap, would be nice to add for later.

src/VideoCommand.cc
luisremis: gosh, not sure how it got there. It will make people think I don't use gdb (which I don't).

src/VideoCommand.cc
luisremis: It is not the same. VCL will run decoding when reading, and in this case we don't need that, we just want the blob to be sent without any decoding. The same is true for images, an inefficiency we , but it is least critical(I think there is an issue already, if not, it's been in my mind for some time already). Decoding video is much more compute intensive.

src/defines.h
luisremis: not sure what you mean. I see them consistent. Look at images for reference.

utils/src/api_schema/api_schema.json
luisremis: nope, this is a result of adding video that we have to split operations block in two.

SConstruct
```diff
@@ -35,27 +35,28 @@ def buildServer(env):
...
                   'src/ImageCommand.cc',
-                  'src/ExceptionsCommand.cc',
                   'src/DescriptorsManager.cc',
                   'src/DescriptorsCommand.cc',
                   'src/BoundingBoxCommand.cc',
+                  'src/VideoCommand.cc',
                 ]
 
-  vdms = env.Program('vdms', vdms_server_files)
+  env.Program('vdms', vdms_server_files)
```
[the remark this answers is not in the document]
luisremis: these are minor indentation issues. don't see a point on making a separate commit for them. Given the file has to be updated, we can fix these minor indentation issues.

src/ImageCommand.cc
```diff
@@ -103,21 +103,21 @@ int AddImage::construct_protobuf(PMGDQuery& query,
     }
 
     std::string img_root = _storage_tdb;
-    VCL::ImageFormat vcl_format = VCL::TDB;
+    VCL::Image::Format vcl_format = VCL::Image::Format::TDB;
```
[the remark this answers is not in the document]
luisremis: I agree. For now we just update the changes in the enums. We can change namespacing in a future refactor, I already have some other things in mind that we can do better.

tests/python/TestImages.py
```diff
@@ -251,6 +251,7 @@ def test_addImageWithLink(self):
         addImage = {}
         addImage["properties"] = props
         addImage["link"] = link
+        addImage["format"] = "png"
```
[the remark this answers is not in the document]
luisremis: yes, I think we do, same for blobs

src/VideoCommand.cc
vishakha041: In that case I would have added a function to VCL to allow the same thing so all data manipulation stayed in one library as much as possible but too late I support. FIle an issue?

SConstruct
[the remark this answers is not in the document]
vishakha041: The ones later in some other scons file looked like more than that. Indentation is fine of course.

src/VideoCommand.cc
vishakha041: I actually prefer the w, h etc that's on the wiki (minus some formatting issues there). Mapping it to current code will make it inconsistent with what bounding box API has. Christina and I had discussed the variable names and she had made them match in code and wiki for that commit.

src/defines.h
vishakha041: Haha. they are consistently inconsistent :)  what's with some caps, some small in the text following :

vishakha041: I think I see a pattern. The tags are all caps but property keys are small after :

src/VideoCommand.cc
luisremis: I see. Doing that for the boundingbox was a mistake, as it broke consistency with what we already had for operations (shame we didn't catch that before). 
Image operations use (and have always used) height and width. Videos use the same operations block.
We should keep them consistent with the images as they are now.

vishakha041: I dont think it was a mistake. A rectangle block with x, y, w, h makes sense. With crop for example, keeping x, y and then width and height is looking weird now.

SConstruct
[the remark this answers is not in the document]
luisremis: will make a separate commit for utils/Sconscript, that makes sense.

PR #76, opened 2019-01-30
Update Dockerfile

luisremis:


docker/Dockerfile
```diff
@@ -1,80 +1,106 @@
...
-RUN rm -rf /*.tar.gz
+# Valijson
+RUN cd / && git clone https://github.com/tristanpenman/valijson.git && \
+    cd valijson && cp -r include/* /usr/local/include/
 
 # VDMS install
-RUN cd / && wget https://github.com/intellabs/vdms/archive/v1.0.0.tar.gz && tar xf v1.0.0.tar.gz
-RUN mv vdms-1.0.0 vdms && cd vdms && mkdir db && scons -j16 INTEL_PATH=/
+RUN git clone https://github.com/intellabs/vdms/ && \
+    cd vdms && git checkout develop && \
```
vishakha041: Will you be changing this to 2.0 checkout? Should this commit go after the merge?

docker/Dockerfile
```diff
@@ -1,80 +1,106 @@
...
 RUN rm -rf /*.tar.gz
 RUN rm -rf /var/lib/apt/lists/* /root/.cache
 
+# Setup entry point
+
 RUN echo '#!/bin/bash' > /start.sh
-RUN echo 'export LD_LIBRARY_PATH=/pmgd/lib:$(find /usr/local/lib/ / -type f -name "*.so" | xargs dirname | sort | uniq | tr "\n" ":")' >> /start.sh
+RUN echo 'export LD_LIBRARY_PATH=/vdms/utils:/pmgd/lib:$(find /usr/local/lib/ / -type f -name "*.so" | xargs dirname | sort | uniq | tr "\n" ":")' >> /start.sh
 RUN echo 'cd /vdms' >> /start.sh
 RUN echo 'rm -rf db' >> /start.sh
```
vishakha041: CAn we add a line here to mkgraph using reasonable params like multiple allocators? Sadly we cannot create one with indexes and such but something. Maybe in the docker with samples

luisremis: this will use the default number of allocators that can be set through vdms config file (we could change our default). Also, users can modify this param and re-run vdms in the instance, if needed.

PR #77, opened 2019-01-30
Remove old, unused files

luisremis:


vishakha041:
Does this also remove the folder these files are in? Those were the only two files in that folder

vishakha041:
There is also still RWLock.h and arch.h files which I think are not needed anymore. YOu can change this one to code cleanup branch and put two commits as far as I can tell right now

luisremis:
Good suggestion. I removed those files as well.

luisremis:
> Does this also remove the folder these files are in? Those were the only two files in that folder

Yes, it does.

PR #81, opened 2019-02-20
Add descriptors metric

luisremis:
Expose the metric param for creation of Descriptor Sets

### DIFF
--- a/utils/src/api_schema/api_schema.json
+++ b/utils/src/api_schema/api_schema.json
@@ -102,6 +102,16 @@
       "enum": ["png", "jpg"]
     },
 
+    "metricFormatString": {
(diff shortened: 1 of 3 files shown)


utils/src/api_schema/api_schema.json
```diff
@@ -534,6 +544,8 @@
       "properties": {
         "name":       { "type": "string" },
         "dimensions": { "$ref": "#/definitions/refInt" },
+        "metric":     { "$ref": "#/definitions/metricFormatString" },
```
vishakha041: Was this already implemented and only the JSON check was missing?

luisremis: yes

PR #93, opened 2019-03-20
Add vcl files and update building/testing files

luisremis:
Integrates vcl, integrates unit tests, updates INSTALL instructions.


tests/SConscript
```diff
@@ -5,25 +5,49 @@ import os
...
+                'unit_tests/DescriptorSetAdd_test.cc',
+                'unit_tests/DescriptorSetClassify_test.cc',
+                'unit_tests/DescriptorSetTrain_test.cc',
+                'unit_tests/DescriptorSetReadFS_test.cc',
+                'unit_tests/DescriptorSetStore_test.cc',
+              ]
 
 query_tests = testenv.Program(
-                'query_tests',
+                'unit_test',
```
vishakha041: I think this commit is very mixed up. If you want to split the tests into unit tests and general tests, then that should go before merging vcl. Also, json_queries.cc is not a unit test. It tests multiple components at the same time. pmgd_queries.cc is a unit test.

SConstruct
```diff
@@ -10,21 +10,27 @@ AddOption('--timing', action='append_const', dest='cflags',
...
+    CPPPATH= ['src', 'include', 'src/vcl', 'utils/include',
               '/usr/include/jsoncpp/',
               os.path.join(env['INTEL_PATH'], 'pmgd/include'),
               os.path.join(env['INTEL_PATH'], 'pmgd/util'),
-              os.path.join(env['INTEL_PATH'], 'vcl/include'),
-              os.path.join(env['INTEL_PATH'], 'vcl/src'),
              ],
     LIBS = [ 'pmgd', 'pmgd-util',
              'jsoncpp', 'protobuf', 'tbb',
-             'vdms-utils', 'vcl', 'pthread',
```
vishakha041: I understand that you do not want to make a library out of the vcl folder because no other application will link against it. But I think library is more of a self contained unit of one kind of functionality. Unless there is a performance reason not to, it might still be sensible to make a library and clean up this scons file. All the opencv stuff will get encapsulated there that way.

vishakha041: Also, by that logic, we will have to refactor utils too.

tests/unit_tests/main.cc
```diff
@@ -0,0 +1,42 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+#include "gtest/gtest.h"
+
+int main(int argc, char **argv)
+{
```
vishakha041: Does it make sense to add the VCL test code in a separate commit?

tests/unit_tests/main.cc
```diff
@@ -0,0 +1,42 @@
...
+    ::testing::InitGoogleTest(&argc, argv);
+
+    // To make GoogleTest silent:
+    // if (true) {
+    //     auto& listeners = ::testing::UnitTest::GetInstance()->listeners();
+    //     delete listeners.Release(listeners.default_result_printer());
```
vishakha041: I see that the config files for unit tests are all still in the main tests folder. Is there a reason for that?

tests/SConscript
vishakha041: Also, I think the organization for tests now needs to segregate tests/unit_tests and tests/server. Then inside server, you can have python and cpp. That will capture the situation better I feel.

luisremis: we will keep the test merge in the same commit, we will separate the tests folder as we discussed.

SConstruct
luisremis: we will keep vcl as a module, but not as a separate library. utils is actually linked against more than one component, vcl is not.

PR #96, opened 2019-03-29
Fix fimg noblob (#90)

### LINKED ISSUE #90: FindImage with blob=False crashes server
FindImage with blob=False crashes server

Server error:
```
"vdms: src/ImageCommand.cc:238: virtual Json::Value VDMS::FindImage::construct_responses(Json::Value&, const Json::Value&, VDMS::protobufs::queryMessage&, const string&): Assertion `findImage.isMember(""entities"")' failed.
/start.sh: line 5:    13 Aborted                 (core dumped) ./vdms
```

luisremis:
Fix #90

### DIFF
--- a/src/ImageCommand.cc
+++ b/src/ImageCommand.cc
@@ -235,84 +235,89 @@ Json::Value FindImage::construct_responses(
 
     Json::Value& findImage = responses[0];
 
-    assert(findImage.isMember("entities"));
-
     if (findImage["status"] != 0) {
         findImage["status"]  = RSCommand::Error;
         // Uses PMGD info error.
         return error(findImage);
     }
 
-    bool flag_empty = true;
+    Json::Value results = get_value<Json::Value>(cmd, "results");
 
-    for (auto& ent : findImage["entities"]) {
+    bool flag_empty = false;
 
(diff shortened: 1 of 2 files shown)


luisremis:
@vishakha041, git gets confused on the changes on ImageCommand.cc. I recommend you look at the files directly to see the changes (that are way less complicated than what it appears)

src/ImageCommand.cc
```diff
@@ -235,85 +235,90 @@ Json::Value FindImage::construct_responses(
...
-    assert(findImage.isMember("entities"));
-
     if (findImage["status"] != 0) {
         findImage["status"]  = RSCommand::Error;
         // Uses PMGD info error.
         return error(findImage);
     }
 
-    bool flag_empty = true;
+    Json::Value results = get_value<Json::Value>(cmd, "results");
```
vishakha041: Actually I am quite confused with this commit. Shouldn't we check in construct_protobuf if blob == false and no list/count/average requested? Or is this relying on PMGDQueryHandler throwing an error and then handling it construct_responses? Ideally we should put it in JSON checks but I am guessing that's not possible. What am I missing?

PR #100, opened 2019-04-22
Remove VCL from Base Docker

luisremis:
All tests passing, new docker images updated at hub.docker.com/intellabs


luisremis:
Yes, it does.

PR #104, opened 2019-05-03
Remove copies in image and video implementation

luisremis:
Optimized video and image code, avoiding copies of cvmats. This increase performance of video read and several image operations by up to 25%.

### DIFF
--- a/include/vcl/Image.h
+++ b/include/vcl/Image.h
@@ -84,7 +84,7 @@ namespace VCL {
          *
          *  @param cv_img  An OpenCV Mat that contains an image
          */
-        Image(const cv::Mat &cv_img);
+        Image(const cv::Mat &cv_img, bool copy=true);
 
         /**
          *  Creates an Image object from an encoded buffer
@@ -112,10 +112,20 @@ namespace VCL {
 
         /**
          *  Creates a new Image object from an existing Image object
+         *  This will make a deep copy of the arrays.
          *
          *  @param img  An existing Image object
          */
         Image(const Image &img);
+
+        /**
+         *  Move constructor, needed to avoid copies of the arrays.
+         *  noexcept is needed to let vectors grow and call the move
+         *  instead of copy constructor.
+         *
+         *  @param img  An existing Image object
+         */
+        Image(const Image &&img) noexcept;
 
         /**
          *  Sets an Image object equal to another Image object
@@ -178,9 +188,11 @@ namespace VCL {
         /**
          *  Gets an OpenCV Mat that contains the image data
          *
+         *  @param copy Specify if a deep copy of the cvmat will be made before
+         *              returning the cvmat object.
          *  @return An OpenCV Mat
          */
-        cv::Mat get_cvmat() const;
+        cv::Mat get_cvmat(bool copy=true);
 
         /**
          *  Gets the raw image data
--- a/src/vcl/ImageData.cc
+++ b/src/vcl/ImageData.cc
@@ -35,7 +35,6 @@
 
 #include "ImageData.h"
 
-
 using namespace VCL;
 
     /*  *********************** */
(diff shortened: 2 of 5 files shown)


vishakha041:
I don't see new tests for checking the copy parameters. Are our current tests sufficient to make sure we are not losing image content anywhere? Our typical checks involve metadata verification. I guess that concern applies to the old code too but just thought of it now.

vishakha041:
What are the pros/cons of merging the ImageData/Image classes? That code was reviewed well back when and this change kind of breaks consistency across how video and descriptors were structured. Is that not true?

luisremis:
> What are the pros/cons of merging the ImageData/Image classes? That code was reviewed well back when and this change kind of breaks consistency across how video and descriptors were structured. Is that not true?

These changes have been moved to another branch and will be part of a future pull request

luisremis:
> I don't see new tests for checking the copy parameters. Are our current tests sufficient to make sure we are not losing image content anywhere? Our typical checks involve metadata verification. I guess that concern applies to the old code too but just thought of it now.

We have tests for both, the metadata an pixels. We have many tests that check every single pixel of the image and video and compare it with direct OpenCV results. The copies in the images are tests through the video implementation that is using those copie/move constructors. All tests are passing.

src/vcl/ImageData.cc
```diff
@@ -318,26 +330,34 @@ ImageData::ImageData(void* buffer, cv::Size dimensions, int cv_type)
...
+        }
+    }
 
     if ( img._tdb != NULL )
         _tdb = new TDBImage(*img._tdb);
     else
         _tdb = NULL;
 
     int start;
+    // TODO IS THIS REALLY NEEDED????
```
vishakha041: Its better to not have comments like that. Either you can resolve it or file an issue and remove the comment

src/vcl/ImageData.cc
```diff
@@ -752,16 +775,28 @@ void ImageData::delete_object()
...
     _channels = cv_img.channels();
 
     _height = cv_img.rows;
-    _width = cv_img.cols;
+    _width  = cv_img.cols;
 
     _cv_type = cv_img.type();
 
-    _cv_img = cv_img.clone();
+    _cv_img = cv_img; // shallow copy
```
vishakha041: Probably a silly question but is this still thread safe? That would be hard to catch with our tests.

src/vcl/VideoData.cc
```diff
@@ -137,7 +137,7 @@ void VideoData::Write::operator()(VideoData *video)
     std::vector<VCL::Image>& frames = video->get_frames();
 
     for (auto& frame : frames) {
-        outputVideo << frame.get_cvmat();
+        outputVideo << frame.get_cvmat(false);
```
vishakha041: I can see why the shallow copy would be important here!

include/vcl/Image.h
```diff
@@ -86,6 +86,8 @@ namespace VCL {
          */
         Image(const cv::Mat &cv_img);
 
+        Image(cv::Mat &cv_img);
```
luisremis: see if we can use a parameter instead

src/vcl/ImageData.cc
```diff
@@ -381,7 +401,8 @@ void ImageData::operator=(const ImageData &img)
         std::shared_ptr<Operation> front = img._operations.front();
         if (front->get_type() == OperationType::READ) {
             start = 1;
-            copy_cv(cv::imread(img._image_id, cv::IMREAD_ANYCOLOR));
+            cv::Mat img_read = cv::imread(img._image_id, cv::IMREAD_ANYCOLOR);
```
luisremis: check if this hurts or not.

src/vcl/ImageData.cc
luisremis: none of this is thread safe. these changes do not affect that.

luisremis:
All comments applied, all tests passing, noticeable improvement on video performance by avoing copies. Same should be good images that is now avoiding multiple copies. looks good to go!

PR #106, opened 2019-05-14
Update INSTALL.md

luisremis:


vishakha041:
Is this for updated versions or packages that are not needed? It might be better to mention the purpose in the commit message. I feel like I saw another PR for install file a couple days back and now I am confused why.

PR #108, opened 2019-05-20
Remove api layers

mahircg:


mahircg:
[without_video_merge.log](https://github.com/IntelLabs/vdms/files/3198665/without_video_merge.log)
[with_image_merge.log](https://github.com/IntelLabs/vdms/files/3198666/with_image_merge.log)
[with_video_merge.log](https://github.com/IntelLabs/vdms/files/3198667/with_video_merge.log)
[without_image_merge.log](https://github.com/IntelLabs/vdms/files/3198668/without_image_merge.log)

I'm also attaching the logs for unit-test execution before and after merging the classes.

src/vcl/Image.cc
```diff
@@ -100,125 +441,444 @@ Image::~Image()
...
 {
-    return _image->get_image_id();
+    return _image_id;
 }
 
-cv::Size Image::get_dimensions() const
+cv::Size Image::get_dimensions()
 {
-    return _image->get_dimensions();
+    //  Previously, this function called perform_operations() in order
```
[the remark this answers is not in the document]
mahircg: Done

include/vcl/Video.h
```diff
@@ -256,10 +480,57 @@ namespace VCL {
      */
     void store(const std::string &video_id, Codec video_codec);
 
+    /**
+     *  Stores a Write Operation in the list of operations to perform.
```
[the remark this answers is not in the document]
mahircg: I extended the comment to reflect the actual behaviour

philiplantz:
The commit message for commit e8cd292 mentions commit ea80b48, which has changed (and will presumably change again). Perhaps it would be better to remove the commit id from the message, and say "a prior commit".

mahircg:
> The commit message for commit [e8cd292](https://github.com/IntelLabs/vdms/commit/e8cd29295dc4bad46c03721e36841a814f0ba61e) mentions commit [ea80b48](https://github.com/IntelLabs/vdms/commit/ea80b48158728f935893b6734e2f3eb12f470294), which has changed (and will presumably change again). Perhaps it would be better to remove the commit id from the message, and say "a prior commit".

Agreed, I'll edit it

philiplantz:
Thanks. I suggest you wait until Vishakha and I have finished our reviews before you make any changes.

Philip

vishakha041:
Are these all the latest commits or are there pending changes based on Luis/Philip's comments? Just want to make sure I review the latest code. Thanks.

mahircg:
> Are these all the latest commits or are there pending changes based on Luis/Philip's comments? Just want to make sure I review the latest code. Thanks.

I haven't done the changes pointed out by Luis/Philip yet. Since those are rather small changes, the code won't change drastically once the changes are applied. I'd say it's good to go with the review.

Thanks

include/vcl/Image.h
```diff
@@ -325,6 +326,432 @@ namespace VCL {
...
+        /**
+         *  Provides a way to keep track of what operations should
+         *   be performed on the data when it is needed
+         *
+         *  Operation is the base class, it keeps track of the format
+         *   of the image data, defines a way to convert Format to
+         *   a string, and defines a virtual function that overloads the
+         *   () operator
+         */
+        class Operation {
```
vishakha041: Since this commit is about simplifying code and removing duplicates, I see a lot of similarities between Operations in images and videos if that were to be a template with specialization. 
In any case, this header is too big and you might consider segregating operations part from the rest. We did that at some point with PMGDIterators since it was getting so complex to contain in one. The class was still a subclass but declarations and definitions were in separate header/cc files.

include/vcl/Image.h
```diff
@@ -213,7 +218,7 @@ namespace VCL {
          *  @see OpenCV documentation for imencode for more details
          */
         std::vector<unsigned char> get_encoded_image(VCL::Image::Format format,
-                const std::vector<int>& params=std::vector<int>()) const;
+                const std::vector<int>& params=std::vector<int>());
```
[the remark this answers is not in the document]
vishakha041: Mainly, if those commits have been merged here, then the removing API layers isn;t the only change here. It also introduced changes to copying methods. So the name / commit messages should reflect that. and preferably be separate. It is hard to isolate problems when commits are combined. I must admit, it is entirely possible that I am seeing a condensed version due to github GUI. So if that is the case, ignore my comment. But if you have combined the shallow/deep copy changes with APi layers into one commit, please segregate them.

src/vcl/Image.cc
```diff
@@ -31,19 +31,261 @@
...
+    /*       READ OPERATION     */
+    /*  *********************** */
+
+Image::Read::Read(const std::string& filename, Image::Format format)
+    : Operation(format),
+      _fullpath(filename)
+{
+}
+
+void Image::Read::operator()(Image *img)
```
vishakha041: See my comment about separating operators in a different file.

src/vcl/Image.cc
```diff
@@ -31,19 +31,261 @@
...
+            img->_tdb->write(_fullpath, _metadata);
+        else
+            img->_tdb->write(img->_cv_img, _metadata);
+    }
+    else {
+        cv::Mat cv_img;
+        if (_old_format == Image::Format::TDB)
+            cv_img = img->_tdb->get_cvmat();
+        else
+            cv_img = img->_cv_img;
```
vishakha041: I am not re-reviewing all this code since we have done it in the past. I am guessing that hasn't changed beyond its move into the up-level class.

vishakha041: Have you seen any performance improvement with this change btw? Just curious

tests/SConscript
```diff
@@ -34,7 +34,6 @@ test_sources = ['main.cc',
                 'unit_tests/pmgd_queries.cc',
                 'unit_tests/helpers.cc',
                 'unit_tests/TDBImage_test.cc',
-                'unit_tests/ImageData_test.cc',
```
vishakha041: Better to not use a commit id in the commit message here unless you are done changing code.

tests/unit_tests/Image_test.cc
```diff
@@ -124,6 +124,32 @@ class ImageTest : public ::testing::Test {
     int size_;
 };
 
+namespace VCL {
+
+    class ImageTest : public Image{
+
```
vishakha041: Were any of the tests failing in the first commit that removed ImageData? If yes, either it would make sense to fix it in the same commit or remove them before hand to avoid a non-working commit OR make a note in the commit message.

include/vcl/Video.h
```diff
@@ -256,10 +261,279 @@ namespace VCL {
...
+        /**
+         *  Provides a way to keep track of what operations should
+         *   be performed on the data when it is needed
+         *
+         *  Operation is the base class, it keeps track of the format
+         *   of the Video data, defines a way to convert Format to
+         *   a string, and defines a virtual function that overloads the
+         *   () operator
+         */
+        class Operation {
```
vishakha041: Similar comment as in Image operations.

tests/SConscript
```diff
@@ -36,7 +36,6 @@ test_sources = ['main.cc',
                 'unit_tests/TDBImage_test.cc',
                 'unit_tests/Image_test.cc',
                 'unit_tests/Video_test.cc',
-                'unit_tests/VideoData_test.cc',
```
vishakha041: Similar comment for using commit ids.

include/vcl/Image.h
```diff
@@ -71,23 +65,34 @@ namespace VCL {
     /*  *********************** */
     /*        CONSTRUCTORS      */
     /*  *********************** */
+
+        /**
+         *  Default constructor, creates an empty Image object.
+         *    Used when reading from the file system
+         */
+        Image();
+
```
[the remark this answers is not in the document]
mahircg: Agreed, I'll make it private.

include/vcl/Image.h
```diff
@@ -71,23 +65,34 @@ namespace VCL {
...
         /**
          *  Creates an Image object from the image id (where the
-         *    image data can be found in the system)
+         *    image data can be found in the system). Note that this
+         *    constructor does not read the image, similar to default
+         *    constructor.
```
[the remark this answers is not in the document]
mahircg: Agreed.

include/vcl/Image.h
```diff
@@ -71,23 +65,34 @@ namespace VCL {
...
          *
          *  @param image_id  The full path to the image
          */
         Image(const std::string &image_id);
 
         /**
-         *  Creates an Image object from the OpenCV Mat
+         *  Creates an Image object from the OpenCV Mat,
+         *  making a deep copy
```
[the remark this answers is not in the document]
mahircg: Agreed (in retrospect, I was tempted to write it as the default parameter is true but that shouldn't really change it).

include/vcl/Image.h
```diff
@@ -98,41 +103,41 @@ namespace VCL {
...
          *  @see OpenCV documentation for more information on type and Size
          */
-        // template <class T> Image(const T* buffer, cv::Size dimensions,
-        Image(void* buffer, cv::Size dimensions,
-            int cv_type);
+        Image(void* buffer, cv::Size dimensions, int cv_type);
 
         /**
          *  Creates a new Image object from an existing Image object
          *  This will make a deep copy of the arrays.
```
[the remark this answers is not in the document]
mahircg: Agreed

SConstruct
```diff
@@ -65,10 +65,8 @@ def buildServer(env):
                   'src/vcl/Exception.cc',
                   'src/vcl/TDBObject.cc',
                   'src/vcl/Image.cc',
-                  'src/vcl/ImageData.cc',
                   'src/vcl/TDBImage.cc',
                   'src/vcl/Video.cc',
-                  'src/vcl/VideoData.cc',
                   'src/vcl/DescriptorSet.cc',
                   'src/vcl/DescriptorSetData.cc',
```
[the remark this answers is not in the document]
mahircg: AFAIK, the initial attempt was only on Image and Video classes. I don't have a good answer to 'why not', except time and priority.

@luisremis Is it planned?

include/vcl/Image.h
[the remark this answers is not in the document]
mahircg: All these changes are done on top of the changes that has been merged by Luis (commit 5184192). Hence, the changes are not related.

In this specific instance, 'const' is removed because 'get_encoded_image' modifies the members of the Image class now that the ImageData class is gone (previously, it was const because no member of the Image class was modified)

include/vcl/Image.h
mahircg: I agree that a template for operations that receive the media (image, video) as type would look good. I suppose given the difficulty in reviewing this huge change set (due to removed/copied lines), we may be better of doing it in another patch.

As a side note, we discussed this as part of a more recent change. The use case is as follows: a frame is extracted from a video as result of 'FindFrame' command, and the command contains the operations of 'Image' type rather than 'Video'. Since the video command class processes and checks (in VideoCommand.c) only video operations, we had to expose Image operations through a dirty hack (as you'll see when reviewing FindFrames patch). Seems like templated operations may solve this very problem.

src/vcl/Image.cc
mahircg: > I am not re-reviewing all this code since we have done it in the past. I am guessing that hasn't changed beyond its move into the up-level class.

I copy-pasted these after Luis' patch, so yes, they're not changed except getting moved.

mahircg: > Have you seen any performance improvement with this change btw? Just curious

Yes, a pretty minuscule amount though. Since I pushed this change almost a month ago, I can't remember the exact improvement. Please check the test logs I attached. They're far from proper benchmarks, but give a rough idea on the execution time change.

tests/SConscript
mahircg: Agreed, I changed the commit id in 67d2399. Just realized I did the same in 5a1e9dc, so I'll fix that too.

tests/unit_tests/Image_test.cc
mahircg: Yes, there are actually build errors due to the presence of private constructors and other functions. Except that, the tests passed fine.

The changes on merging the classes (f3966e6: Merge ImageData and Image classes) and merging the unit-tests (e8cd292: Merge unit ...)  are done separately, so there'll be build errors if f3966e6 is applied stand-alone. Reason for this was to not make changes TestImageData_test class, which would be removed in e8cd292 anyway. 

Fixing the errors on the same commit will result in writing code that'll be removed in the next patch (since some of the tests are removed when merging ImageData and Image test classes, as they're duplicates).

I guess the best would be to squash those three commits into one, but that'd make the change even bigger. Not sure what's the best approach here.

tests/SConscript
mahircg: Agreed, will fix it.

include/vcl/Video.h
mahircg: Ditto for Image operations: I suggest we do that in a separate patch.

src/vcl/Image.cc
mahircg: I suggest creating a separate patch for it, as this set got already big enough.

tests/unit_tests/Image_test.cc
philiplantz: I think we have three choices (as Vishakha indicated).
0) What we have now: some tests are fixed in the inital commit, but some other fail.
1) Don't make any changes to the tests in the initial commit; add a note to the commit message that tests will fail until the next commit. This has long-term affects if we ever want to do a git-bisect.
2) Make sure that all the tests pass. I guess this effectively means merging all three commits.
3) Remove the ImageData tests in a commit prior to the commit that removes ImageData. In the commit that merges Image and ImageData, make the minimal changes to the tests so that they still pass. In a subsequent commit, add the new Image tests to test the "new" functions in the Image class.

mahircg: If we go with 1st option, we'll have build errors due to constructors/functions that became private members as result of the merging. Furthermore, as you mentioned, hitting that commit as result of a git-bisect will be unpleasant. Let me take care of the build errors (by applying the 'friend' class trick, just to be removed in next patch). If the tests pass, we should be good. The only downside is I'll write code that's deleted just in the next patch, but I guess it's pretty much the same effort with the 3rd option.

2nd option looks the easiest. I can squash  f3966e6 and e8cd292, and rebase the move constructor change on the squashed changes. The change will be huge, though.

include/vcl/Image.h
vishakha041: Separate patch after this merge is fine. It will be nicely contained and still be part of the code cleanup sort of theme.

tests/unit_tests/Image_test.cc
vishakha041: Seems like the option 3 might be the cleanest and easiest. That way you are not changing tests that get deleted. Just adding new test cases after the functionality. Is that not true? And none of your commits will break most likely.

mahircg: Agreed, let's do the third way. I've already tested the 1st option, and even though there are no UT errors once the build errors are resolved, the 3rd option seems cleaner as we do not add any lines that are removed just in next patch.

There may be merge conflicts, but shouldn't be much of an issue. I'll apply the 3rd option and push the changes.

philiplantz:
Vishakha and I had both commented that any API changes to the Image class should be separate from the Image/ImageData merge. But I had missed the fact that the body of the ImageData constructor was incorporated into Image as part of the merge and that it already had the copy parameter on the constructor, so that capability naturally got pulled in.
The only other API change was the addition of the default constructor, which has been made private, so it isn't actually an API change.
So I think I'm now okay with not making a separate commit for API changes.

luisremis:
I can't respond to this directly for some reason: https://github.com/IntelLabs/vdms/pull/108#discussion_r297830471 

We will do a separate patch for that, not really a good reason not to other than video and images are high priority now. I will remove the separation on the feature vector implementation later and also add fixes that I have in my local branches as well.

vishakha041:
I don't think github will send an email if you update the branch here. So please let us know whenever you are done so we can approve the PR. It should be good since you have applied all the changes.

SConstruct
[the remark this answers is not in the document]
luisremis: we will leave the descriptors layering fix for a separate PR.

SConstruct
```diff
@@ -67,7 +67,6 @@ def buildServer(env):
                   'src/vcl/Image.cc',
                   'src/vcl/TDBImage.cc',
                   'src/vcl/Video.cc',
-                  'src/vcl/VideoData.cc',
```
vishakha041: Did you not face the VideoData test problem here?

tests/SConscript
vishakha041: The commit message here still has a commit Id

SConstruct
luisremis: Nope, because video testing is more on interface methods only.

luisremis:
Fixed the last commit message. Good to go!

PR #110, opened 2019-06-21
Add FindFrames command

mahircg:
This patch introduces 'FindFrames' command, that searches for the given
video name and returns a list of frames from the video if it is found.

Currently, FindFrame command relies on OpenCV's frame iteration
capabilities. Thus, the entire video is loaded into memory to be
decoded. In future, the video loading and decoding should be done only
on the requested frame interval.

### DIFF
--- a/src/VideoCommand.cc
+++ b/src/VideoCommand.cc
@@ -32,6 +32,7 @@
 #include <iostream>
 #include <fstream>
 
+#include "ImageCommand.h" // for enqueue_operations of Image type
 #include "VideoCommand.h"
 #include "VDMSConfig.h"
 #include "defines.h"
@@ -95,6 +96,26 @@ VCL::Video::Codec VideoCommand::string_to_codec(const std::string& codec)
     return VCL::Video::Codec::NOCODEC;
 }
 
+Json::Value VideoCommand::check_responses(Json::Value& responses)
+{
+    if (responses.size() != 1) {
+        Json::Value return_error;
+        return_error["status"]  = RSCommand::Error;
+        return_error["info"] = "PMGD Response Bad Size";
+        return return_error;
+    }
+
+    Json::Value& response = responses[0];
+
+    if (response["status"] != 0) {
+        response["status"]  = RSCommand::Error;
+        // Uses PMGD info error.
+        return response;
+    }
+
+    return response;
+}
+
 //========= AddVideo definitions =========
 
 AddVideo::AddVideo() : VideoCommand("AddVideo")
@@ -152,7 +173,19 @@ int AddVideo::construct_protobuf(
     return 0;
 }
 
-//========= UpdateImage definitions =========
+Json::Value AddVideo::construct_responses(
+    Json::Value& response,
+    const Json::Value& json,
+    protobufs::queryMessage &query_res,
+    const std::string& blob)
+{
+    Json::Value ret;
+    ret[_cmd_name] = RSCommand::check_responses(response);
+
+    return ret;
+}
+
+//========= UpdateVideo definitions =========
 
 UpdateVideo::UpdateVideo() : VideoCommand("UpdateVideo")
 {
@@ -251,22 +284,12 @@ Json::Value FindVideo::construct_responses(
         return ret;
     };
 
-    if (responses.size() != 1) {
-        Json::Value return_error;
-        return_error["status"]  = RSCommand::Error;
-        return_error["info"] = "PMGD Response Bad Size";
-        error(return_error);
+    Json::Value resp = check_responses(responses);
+    if (resp["status"] != RSCommand::Success) {
+        return error(resp);
     }
 
     Json::Value& FindVideo = responses[0];
-
-    assert(FindVideo.isMember("entities"));
-
-    if (FindVideo["status"] != 0) {
-        FindVideo["status"]  = RSCommand::Error;
-        // Uses PMGD info error.
-        error(FindVideo);
-    }
 
     bool flag_empty = true;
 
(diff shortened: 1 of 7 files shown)


vishakha041:
I'll look at this once we close the other PR. I believe this is based off of that one.

luisremis:
@vishakha041, these are independent, as you can see following the commits. So you can go ahead and review it.

src/VideoCommand.cc
```diff
@@ -243,31 +268,16 @@ Json::Value FindVideo::construct_responses(
 {
     const Json::Value& cmd = json[_cmd_name];
 
-    Json::Value ret;
+    Json::Value ret = check_responses(responses);
```
vishakha041: Shouldn't you return from here if there was an error? I suppose you have introduced this function to avoid repeating code below but the original code is to stop executing on error which I am not sure this one does. Also, if you haven't already, please do add failure test cases too.

src/VideoCommand.cc
```diff
@@ -350,3 +360,194 @@ Json::Value FindVideo::construct_responses(
...
+    PMGDQuery& query,
+    const Json::Value& jsoncmd,
+    const std::string& blob,
+    int grp_id,
+    Json::Value& error)
+{
+    const Json::Value& cmd = jsoncmd[_cmd_name];
+
+    // We try to catch the missing attribute error before
+    // initiating a PMGD query
```
vishakha041: Good idea!

src/VideoCommand.cc
```diff
@@ -350,3 +360,194 @@ Json::Value FindVideo::construct_responses(
...
+
+Json::Value FindFrames::construct_responses(
+    Json::Value& responses,
+    const Json::Value& json,
+    protobufs::queryMessage &query_res,
+    const std::string &blob)
+{
+    const Json::Value& cmd = json[_cmd_name];
+
+    Json::Value ret = check_responses(responses);
```
vishakha041: Same comment as above

src/VideoCommand.cc
```diff
@@ -350,3 +360,194 @@ Json::Value FindVideo::construct_responses(
...
+    };
+
+    Json::Value& FindFrames = responses[0];
+
+    bool flag_empty = true;
+
+    for (auto& ent : FindFrames["entities"]) {
+
+        if(!ent.isMember(VDMS_VID_PATH_PROP)){
+            continue;
```
vishakha041: I think this should be an assert or exception. Isn't this an internal property that must be added when a video goes into the system?

src/VideoCommand.cc
```diff
@@ -350,3 +360,194 @@ Json::Value FindVideo::construct_responses(
...
+{
+    const Json::Value& cmd = jsoncmd[_cmd_name];
+
+    // We try to catch the missing attribute error before
+    // initiating a PMGD query
+    Json::ArrayIndex tmp;
+    bool is_interval = get_interval_index(cmd, tmp);
+    bool is_frames   = cmd.isMember("frames");
+
+    if (!(is_frames != is_interval)) {
```
vishakha041: Can you not request the entire video as frames in which case interval could be absent?
I don't follow the logic here of comparing these variables especially with a ! of != instead of ==

vishakha041: Shouldn't you instead check if start/stop/step are valid variables and start < stop etc.?

src/VideoCommand.cc
```diff
@@ -350,3 +360,194 @@ Json::Value FindVideo::construct_responses(
...
+                for (auto& fr : cmd["frames"]) {
+                    frames.push_back(fr.asUInt());
+                }
+            }
+            else if (is_interval) {
+                Json::Value interval_op = operations[interval_idx];
+                int i;
+                for (
+                    i =  get_value<int>(interval_op, "start");
+                    i <  get_value<int>(interval_op, "stop");
```
vishakha041: I think it might make sense to read the JSON variables before the for loop. Philip has made me less trusting of the compiler over the years :)

src/VideoCommand.cc
```diff
@@ -350,3 +360,194 @@ Json::Value FindVideo::construct_responses(
...
+                    i += get_value<int>(interval_op, "step"))
+                {
+                    frames.push_back(i);
+                }
+                Json::Value deleted;
+                operations.removeIndex(interval_idx, &deleted);
```
vishakha041: What does this do?

src/VideoCommand.cc
```diff
@@ -350,3 +360,194 @@ Json::Value FindVideo::construct_responses(
...
+                return_error["status"]  = RSCommand::Error;
+                return_error["info"] = "No 'frames' or 'interval' parameter";
+                return error(return_error);
+            }
+
+            VCL::Video video(video_path);
+            FindImage  image_op;
+            VCL::Image::Format format = image_op.get_requested_format(cmd);
+
+            if (format == VCL::Image::Format::NONE_IMAGE) {
```
vishakha041: I would call it NO_IMAGE or IMAGE_NONE.

src/VideoCommand.cc
```diff
@@ -350,3 +360,194 @@ Json::Value FindVideo::construct_responses(
...
+            }
+
+
+            for (auto idx : frames) {
+                cv::Mat mat = video.get_frame(idx);
+                VCL::Image img(mat, false);
+                if (!operations.empty()) {
+                    // This is a dirty hack to use enqueue_operations() member
+                    // of ImageCommand class.
+                    // TODO: Check if this can be done in a better way
```
vishakha041: I suggest either you can find a better way or remove the comment.

tests/python/TestVideos.py
```diff
@@ -130,6 +130,136 @@ def test_findVideo(self):
         for i in range(0, number_of_inserts):
             self.assertEqual(response[i]["FindVideo"]["status"], 0)
 
+    def test_FindFramesByFrames(self):
+
+        db = self.create_connection()
```
vishakha041: Please add tests in a separate commit.

utils/src/api_schema/api_schema.json
```diff
@@ -664,6 +694,20 @@
...
       },
 
+      "additionalProperties": false
+    },
+
+    "FindFrames": {
+      "properties": {
+        "frames":      { "$ref": "#/definitions/nonNegativeIntArray" },
+        "_ref":        { "$ref": "#/definitions/refInt" },
+        "link":        { "$ref": "#/definitions/blockLink" },
```
vishakha041: Is there a wiki page for this? Its good to make this definition consistent with the wiki and with other similar commands

src/VideoCommand.cc
luisremis: good point, but for a separate PR (this comes from the image class, which is not touched here).

utils/src/api_schema/api_schema.json
luisremis: https://github.com/IntelLabs/vdms/wiki/FindFrames

src/VideoCommand.cc
luisremis: This checks that only one_of(interval, frames) is specified. If you just use ==, you need to add an extra condition to make sure at least one exist. The condition as is both checks that at least one is set and that only one is set. 

We can also add checks for start/stop/.

luisremis: About start stop:

actually i would not add a control for that, because if a user wants to the frames and properties, and wrongly set start > stop, we will just return an error when we could have returned properties + empty video. 

This may be a condition we can add in both FindFrames and FindVideo in a future PR. For now we keep the behavior the same as FindVideo.

src/VideoCommand.cc
luisremis: removes the interal operation, so that the operations can be passed to the independent frames later(line 516)

luisremis:
All comments addressed. Good to go for me!

mahircg:
I've just did an interactive rebase on all three patches to sign them off. All the changes look good to me.

luisremis:
Created a separate branch for fixing a bug in Video command response. 

Should be good to go.

src/VideoCommand.cc
```diff
@@ -339,7 +339,7 @@ Json::Value FindVideo::construct_responses(
             Json::Value return_error;
             return_error["status"]  = RSCommand::Error;
             return_error["info"] = "VCL Exception";
-            error(return_error);
+            return error(return_error);
```
vishakha041: Probably not the best location for this commit but if this is the only bug fix in the vicinity, then ok.

src/VideoCommand.cc
```diff
@@ -350,3 +373,202 @@ Json::Value FindVideo::construct_responses(
...
+            }
+            else if (is_interval) {
+
+                Json::Value interval_op = operations[interval_idx];
+
+                int start = get_value<int>(interval_op, "start");
+                int stop  = get_value<int>(interval_op, "stop");
+                int step  = get_value<int>(interval_op, "step");
+
+                for (int i = start; i < stop; i += step)  {
```
vishakha041: Does opencv read frames from index 0 or 1? User seems to provide interval from 1 to n and I don't see an adjustment anywhere. Will this miss a frame?

src/VideoCommand.cc
```diff
@@ -350,3 +373,202 @@ Json::Value FindVideo::construct_responses(
...
+                    frames.push_back(i);
+                }
+
+                Json::Value deleted;
+                operations.removeIndex(interval_idx, &deleted);
+            }
+            else {
+                // This should never happen, as we check this condition in
+                // FindFrames::construct_protobuf(). In case this happens, it
+                // is better to signal it rather than to continue
```
vishakha041: If this cannot happen, shouldn't this be an assert since it sounds more like a code problem.

src/VideoCommand.cc
luisremis: It is 0-indexed. The implementation is correct and the api_schema correctly defined. I just updated the example on the wiki to make sure it is clear.

src/VideoCommand.cc
luisremis: It is better to not have assertions and doing it this way as it will log the problem rather than killing the server. Killing the server with an assertion is very painful when writing applications. It is more robust to return an error and log the problem, and we have the comment that explains what can be happening for debug. We need to work in the future on better logging strategies as well.

vishakha041: Depends on whether you are trying to address a user created problem or problem with the server code. 
Agree with the logging comment.

PR #111, opened 2019-06-27
Change default Image format

luisremis:
TDB format only performs better with very large images, and it is usually not the first choice for users, forcing them to explicitly change for format for every insertion or getting error messages. It is better to just go with JPG as the default until we start seeing more use cases showing the benefits of the format.

### DIFF
--- a/src/ImageCommand.cc
+++ b/src/ImageCommand.cc
@@ -103,7 +103,7 @@ int AddImage::construct_protobuf(PMGDQuery& query,
     }
 
     std::string img_root = _storage_tdb;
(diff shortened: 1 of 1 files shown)


vishakha041:
Only small concern here, shouldn't png be default since it is lossless? Or more like, set it to whatever the original is??

luisremis:
sounds good, we can switch to use the original format that the user set, as a feature enhancement, if the use does not specify a change of format nor operations, we need to write the image directly. 
The code as is is always running decoding over the image (some times unnecessarily).

PR #112, opened 2019-06-28
Fix bug in responses for DescriptorsCommand

luisremis:


src/DescriptorsCommand.cc
```diff
@@ -163,7 +163,7 @@ Json::Value AddDescriptorSet::construct_responses(
     };
 
     if (resp["status"] != RSCommand::Success) {
-        error(resp);
+        return error(resp);
```
vishakha041: Aah this one. I had caught it and forgotten to file a bug request last week. I think there was another spot where there was no return.

vishakha041: Can't find it anymore. But just scan through one more time. Wherever check_responses is used.

luisremis: You probably remember this on the remove_api_layer PR, where this was a bug too but I already fixed.

PR #113, opened 2019-07-05
Remove ImageData from unit-test build script

mahircg:
Build script of the unit-test still includes ImageData as dependency.
This causes linking process for unit-tests to fail.

ImageData class has been removed in a previous patch, so it is no longer
needed. This patch removes the dependency.


mahircg:
Seems like I forgot to remove the ImageData from the list of dependencies in the unit-test scons script.
The build did not fail, as the object file was present (not removed by scons -c, as the source file removed). The remnant of the previous build prevented the linker from failing.

PR #114, opened 2019-07-11
Add descriptor blob when using link, Remove entities in results when …

luisremis:
Enables getting a descriptor blob when using FindDescriptor with only link (Case 3 in the FindDescriptor call). Also removes entities in results block when empty.

### DIFF
--- a/src/DescriptorsCommand.cc
+++ b/src/DescriptorsCommand.cc
@@ -518,10 +518,6 @@ int FindDescriptor::construct_protobuf(
     // Add label/id as required.
     // Remove the variables with "_"
     if (results.isMember("list")) {
-
-        results["list"].append(VDMS_DESC_LABEL_PROP);
-        results["list"].append(VDMS_DESC_ID_PROP);
-
         int pos = -1;
         for (int i = 0; i <  results["list"].size(); ++i) {
             if (results["list"][i].asString() == "_label" ||
@@ -534,6 +530,9 @@ int FindDescriptor::construct_protobuf(
             }
         }
     }
+
+    results["list"].append(VDMS_DESC_LABEL_PROP);
+    results["list"].append(VDMS_DESC_ID_PROP);
 
     // Case (1)
     if (cmd.isMember("link")) {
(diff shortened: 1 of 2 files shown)


src/DescriptorsCommand.cc
```diff
@@ -535,6 +531,9 @@ int FindDescriptor::construct_protobuf(
         }
     }
 
+    results["list"].append(VDMS_DESC_LABEL_PROP);
```
vishakha041: Am I right in seeing this as a cosmetic change?

src/DescriptorsCommand.cc
```diff
@@ -934,6 +968,15 @@ Json::Value FindDescriptor::construct_responses(
         }
     }
 
+    if (findDesc.isMember("entities")) {
+        for (auto& ent : findDesc["entities"]) {
+            if (ent.getMemberNames().size() == 0) {
+                findDesc.removeMember("entities");
```
vishakha041: I think for standard return, you should leave entities with a blank [] like it returns when FindEntity has no matching members. It gets very difficult to do standard parsing of results when the behavior is so different across scenarios.

src/DescriptorsCommand.cc
```diff
@@ -735,16 +734,51 @@ Json::Value FindDescriptor::construct_responses(
...
+
+        // These properties should always exist
+        assert(set.isMember(VDMS_DESC_SET_PATH_PROP));
+        assert(set.isMember(VDMS_DESC_SET_DIM_PROP));
+        std::string set_path = set[VDMS_DESC_SET_PATH_PROP].asString();
+        int dim = set[VDMS_DESC_SET_DIM_PROP].asInt();
+
+        if (findDesc.isMember("entities")) {
+
+            if (get_value<bool>(results, "blob", false)) {
```
vishakha041: Isn't this if {} block triplicated in the three conditions? Good time to consolidate?

src/DescriptorsCommand.cc
luisremis: not really. is needed independent if the user specifies results with list now.

src/DescriptorsCommand.cc
luisremis: Not sure, given that re always return a "returned" indicating the number of elements present, and can always be checked for > 0, not to mention that any parse will tell you whether the entities exist or not. For now we keep it consistent with the rest of the commands (they all return no entities block if no property is returned).

luisremis:
all comments addressed.

src/DescriptorsCommand.cc
vishakha041: [
    {
        "FindEntity": {
            "entities": [],
            "returned": 0,
            "status": 0
        }
    }
]

luisremis: I see the confusion. When this type of response was fixed for FindEntity, it was inconsistent with the rest (FindImage, FindVideos, etc). 

For now we keep it consistent with FindImage/FindVideo (they all return no entities block if no property is returned), and we later change them all to match FindEntity of the other way around.

src/DescriptorsCommand.cc
```diff
@@ -668,6 +667,29 @@ int FindDescriptor::construct_protobuf(
...
+                                    const Json::Value& results,
+                                    Json::Value& entities,
+                                    protobufs::queryMessage &query_res)
+{
+    if (get_value<bool>(results, "blob", false)) {
+
+        VCL::DescriptorSet* set = _dm->get_descriptors_handler(set_path);
+        int dim = set->get_dimensions();
+
+        for (auto& ent : entities) {
```
vishakha041: Isn't this separation making you traverse the entity list twice? Once for properties and once for blobs. WOuld it work if you put the loop around entities outside the convert_properties and populate_blobs functions?

luisremis: no, the separation keeps the logic as is. The list is traversed twice in any case because the convert_properties functions aim to trim some of the results if needed as a last step. I am not sure we want to couple the logic for insertion the blobs to the logic to prune/convert the name of some properties. I couldn't think in a way of doing this without looking horribly coupled.

PR #116, opened 2019-07-16
Feature filter on connections

luisremis:
Provides the ability to specify "constraints" in a link block, which enables filtering on edges by its properties.

### DIFF
--- a/src/PMGDQueryHandler.cc
+++ b/src/PMGDQueryHandler.cc
@@ -491,7 +491,15 @@ int PMGDQueryHandler::query_node(const protobufs::QueryNode &qn,
     for (int i = 0; i < qc.predicates_size(); ++i) {
         const PMGDPropPred &p_pp = qc.predicates(i);
         PropertyPredicate j_pp = construct_search_term(p_pp);
-        search.add(j_pp);
+        search.add_node_predicate(j_pp);
+    }
+
+    if (has_link) { // Check for edges constraints
+        for (int i = 0; i < qn.link().predicates_size(); ++i) {
+            const PMGDPropPred &p_pp = qn.link().predicates(i);
+            PropertyPredicate j_pp = construct_search_term(p_pp);
+            search.add_edge_predicate(j_pp);
+        }
     }
 
     PMGD::NodeIterator ni = has_link ?
@@ -609,7 +617,7 @@ int PMGDQueryHandler::query_edge(const protobufs::QueryEdge &qe,
     for (int i = 0; i < qc.predicates_size(); ++i) {
         const PMGDPropPred &p_pp = qc.predicates(i);
         PropertyPredicate j_pp = construct_search_term(p_pp);
-        search.add(j_pp);
+        search.add_node_predicate(j_pp);
     }
 
     EdgeIterator ei = PMGD::EdgeIterator(new NodeEdgeIteratorImpl(search, src_ni, dest_ni));
--- a/src/SearchExpression.cc
+++ b/src/SearchExpression.cc
@@ -58,8 +58,8 @@ class SearchExpression::NodeAndIteratorImpl : public PMGD::NodeIteratorImplIntf
         for (; mNodeIt; mNodeIt.next()) {
             if (_neighbor && (_expr.tag() != 0 && mNodeIt->get_tag() != _expr.tag()) )
                 goto continueNodeIt;
-            for (std::size_t i = _start_at; i < _expr._predicates.size(); i++) {
-                PMGD::PropertyFilter<PMGD::Node> pf(_expr._predicates.at(i));
+            for (std::size_t i = _start_at; i < _expr._node_predicates.size(); i++) {
+                PMGD::PropertyFilter<PMGD::Node> pf(_expr._node_predicates.at(i));
                 if (pf(*mNodeIt) == PMGD::DontPass)
                     goto continueNodeIt;
             }
@@ -77,8 +77,8 @@ public:
     NodeAndIteratorImpl(const SearchExpression &expr)
         : _expr(expr),
           mNodeIt(_expr._db.get_nodes(_expr.tag(),
-                      (_expr._predicates.empty() ? PMGD::PropertyPredicate()
-                           : _expr._predicates.at(0)))),
+                      (_expr._node_predicates.empty() ? PMGD::PropertyPredicate()
+                           : _expr._node_predicates.at(0)))),
           _neighbor(false)
     {
         _start_at = 1;
@@ -93,7 +93,8 @@ public:
                                PMGD::StringID edgetag, bool unique,
                                const SearchExpression &neighbor_expr)
         : _expr(neighbor_expr),
-          mNodeIt(get_neighbors(node, dir, edgetag, unique)),
+          mNodeIt(get_neighbors(node, dir, edgetag,
+                                _expr.get_edge_predicates(), unique)),
           _neighbor(true)
     {
         _start_at = 0;
@@ -135,10 +136,10 @@ class SearchExpression::NodeOrIteratorImpl : public PMGD::NodeIteratorImplIntf
     /// candidate
     bool _next()
     {
-        while (_idx < _expr._predicates.size()) {
+        while (_idx < _expr._node_predicates.size()) {
              PMGD::NodeIterator ni =
                     _expr._db.get_nodes(_expr.tag(),
-                                        _expr._predicates.at(_idx++));
+                                        _expr._node_predicates.at(_idx++));
 
             if (ni) {
                 _node = &*ni;
@@ -153,7 +154,7 @@ class SearchExpression::NodeOrIteratorImpl : public PMGD::NodeIteratorImplIntf
     {
         static int id = 0;
         while (_neighborIt) {
-            for (const auto& pred : _expr._predicates) {
+            for (const auto& pred : _expr._node_predicates) {
                 PMGD::PropertyFilter<PMGD::Node> pf(pred);
                 if (pf(*_neighborIt) == PMGD::Pass) {
                     _node = &*_neighborIt;
@@ -189,7 +190,8 @@ public:
                        PMGD::StringID edgetag, bool unique,
                        const SearchExpression &neighbor_expr)
         : _expr(neighbor_expr),
-          _neighborIt(get_neighbors(node, dir, edgetag, unique)),
+          _neighborIt(get_neighbors(node, dir, edgetag,
+                                    _expr.get_edge_predicates(), unique)),
           _neighbor(true)
     {
         _next_neighbor();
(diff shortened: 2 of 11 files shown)


tests/python/run_python_tests.sh
```diff
@@ -28,6 +28,7 @@ rm log.log screen.log
 rm -r test_db
 
 ../../vdms -cfg config-tests.json > screen.log 2> log.log &
+sleep 1
```
philiplantz: We should investigate a reliable way to do this syncronization.

src/SearchExpression.h
```diff
@@ -62,6 +62,9 @@ namespace VDMS {
         /// The conjunctions of property predicates
         std::vector<PMGD::PropertyPredicate> _predicates;
 
+        /// The conjunctions of property predicates for edges
+        std::vector<PMGD::PropertyPredicate> _e_predicates;
+
```
philiplantz: I suggest renaming these to be more clear: _predicates -> _node_predicates, and _e_predicates -> _edge_predicates.

src/SearchExpression.h
```diff
@@ -71,11 +74,14 @@ namespace VDMS {
             _db(db), _tag(tag), _or(p_or) {}
 
         void add(PMGD::PropertyPredicate pp) { _predicates.push_back(pp); }
+        void add_e(PMGD::PropertyPredicate pp) { _e_predicates.push_back(pp); }
```
philiplantz: As above, I would rename these: add -> add_node_predicate, add_e -> add_edge_predicate

src/SearchExpression.h
```diff
@@ -71,11 +74,14 @@ namespace VDMS {
...
         const PMGD::StringID tag() const { return _tag; };
         PMGD::Graph &db() const { return _db; }
         const PMGD::PropertyPredicate &predicate(int i) const { return _predicates.at(i); }
         const size_t num_predicates() const { return _predicates.size(); }
 
+        std::vector<PMGD::PropertyPredicate> get_e_pp() const { return _e_predicates; }
+
```
philiplantz: Names should be parallel:
get_node_predicate(int i)
num_node_predicates()
get_edge_predicates()

Also for parallelism, it seems you might want to add:
get_node_predicates()
num_edge_predicates()
get_edge_predicate(int i)
but I guess the usage is such that you don't need these, so perhaps it's a foolish consistency.

philiplantz: Shouldn't get_e_pp return a const reference rather than a copy of the vector?

src/QueryHandler.cc
```diff
@@ -368,13 +370,32 @@ void QueryHandler::process_query(protobufs::queryMessage& proto_query,
 
         Json::Value& tx_responses = pmgd_query.run();
 
-        if (tx_responses.size() != root.size()) { // error
+        if (!tx_responses.isArray()) {
+            Json::StyledWriter writer;
+            std::cerr << "PMGD Response:" << std::endl;
+            std::cerr << writer.write(tx_responses) << std::endl;
```
philiplantz: Is this debug trace output that shouldn't be committed?

src/QueryHandler.cc
```diff
@@ -368,13 +370,32 @@ void QueryHandler::process_query(protobufs::queryMessage& proto_query,
...
+
             cmd_result["status"] = RSCommand::Error;
-            Json::StyledWriter w;
-            std::cerr << w.write(tx_responses);
+            cmd_result["info"] = "Failed PMGD Transaction: Wrong Response Size";
+
+            cmd_current = "Transaction";
```
philiplantz: There seems to be a lot of commonality between the two error flows. Could they be combined, using
`if (!tx_responses.isArray() || tx_responses.size() != root.size())`

src/SearchExpression.h
luisremis: I like consistency too, but I think it is better to signal that those are not needed by not putting them there.

src/QueryHandler.cc
luisremis: We have these as logging all over the query handling. We have pending a real logging mechanism.

luisremis:
All comments addressed.

src/PMGDQueryHandler.cc
```diff
@@ -188,7 +188,7 @@ int PMGDQueryHandler::add_node(const protobufs::AddNode &cn,
 {
     long id = cn.identifier();
     if (id >= 0 && _cached_nodes.find(id) != _cached_nodes.end()) {
-        set_response(response, PMGDCmdResponse::Error, "Reuse of _ref value\n");
+        set_response(response, PMGDCmdResponse::Error, "Reuse of _ref value");
```
vishakha041: While this commit is overall necessary, I don't see a relation to the branch at all. I suggest pulling this one and the return message fixing in a separate branch for cleanup.

src/QueryHandler.cc
```diff
@@ -368,13 +370,20 @@ void QueryHandler::process_query(protobufs::queryMessage& proto_query,
...
+            if (!tx_responses.isArray() && tx_responses.isMember("info")) {
+                tx_error_msg += ": " + tx_responses["info"].asString();
+            }
+
             cmd_result["status"] = RSCommand::Error;
-            Json::StyledWriter w;
-            std::cerr << w.write(tx_responses);
+            cmd_result["info"] = tx_error_msg;
+
+            cmd_current = "Transaction";
```
vishakha041: Replacing the failing command name with Transaction doesnt still help find which command failed. There is a wider change for consistency in returning error but even if you want to keep this, please shift this to another branch with the \n commit.

utils/src/protobuf/pmgdMessages.proto
```diff
@@ -207,6 +207,11 @@ message LinkInfo
     // are unique. Does carry a performance penalty to do that
     // in cases where it is known that neighbors are unique.
     bool nb_unique = 14;
+
+    // Specify edge constraints
+    // Specify search conditions and whether to OR or AND them
+    PredicateOp p_op = 20;
+    repeated PropertyPredicate predicates = 21;
```
vishakha041: This branch overall makes me concerned about breaking the logic we had behind our API. We had tried to define link and other blocks to be common pieces to build commands. With this predicate op introduction here, it is an extra thing when you do AddEntity with a link, predicate op makes no sense there. properties would make sense in that scenario. But in order to avoid such specializations we had said, to do more complex things, use FindConnection and the link block was meant to stay for simple scenarios when your edges would not have too many things.

vishakha041: I would have tried to achieve the goal you are going after, by making sure that FindEntity could understand the endpoints of FindConnection and FindConnection iterator had a way to first look at srcni and then predicates, instead of putting constraint on link block directly. IT probably would have amounted to fewer changes overall and not introduced nesting in the blocks.

vishakha041: Since FindConnection already applies property filter, query_node() would need to add a condition to check if ref belongs in _cached_edges() and basically do a get_source() or get_destination() with edge traversal in there. So a new build_results with edge iterator looking at source/destination node.

vishakha041: We can discuss this on the phone if you want.  Perhaps I am missing something in my understanding as to why this way.

src/SearchExpression.cc
```diff
@@ -93,7 +93,8 @@ class SearchExpression::NodeAndIteratorImpl : public PMGD::NodeIteratorImplIntf
                                PMGD::StringID edgetag, bool unique,
                                const SearchExpression &neighbor_expr)
         : _expr(neighbor_expr),
-          mNodeIt(get_neighbors(node, dir, edgetag, unique)),
+          mNodeIt(get_neighbors(node, dir, edgetag,
```
vishakha041: Our previous branches, we have split the PMGD QH level changes from the JSON level changes with unit tests at both levels. I suggest keeping to that. Helps identify breaking points in future commits.

vishakha041: This seems to depend on PMGD changes. The commit should indicate that.

tests/python/TestConnections.py
```diff
@@ -0,0 +1,463 @@
...
+        }
+        all_queries.append(fE)
+
+        fE = {
+            "FindEntity": {
+                "class": "felcflo_foo",
+                "link": {
+                    "ref": 1,
+                    "constraints": {
+                        "probablity": [">=", 0.1],
```
vishakha041: JSON is ugly as it is. I really don't like this nesting. ANd it beats your original point about a simple way to use link.

utils/src/protobuf/pmgdMessages.proto
philiplantz: You're right about the alternatives. Luis and I have discussed them and we should have that same discussion again with you.

src/SearchExpression.cc
luisremis: will skip testing at this level, given that I am already adding tests at VDMS and PMGD levels.

tests/python/run_python_tests.sh
luisremis: good point. Done.

tests/python/TestCommand.py
```diff
@@ -41,6 +41,26 @@ def __init__(self, *args, **kwargs):
...
 
+        db_up = False
+        attempts = 0
+        while(not db_up):
+            try:
+                db = vdms.vdms()
+                db.connect(self.hostname, self.port)
+                db.disconnect()
+                db_up = True
+                if (attempts > 0):
```
mahircg: This won't print the message if the connection is successful in first try.

Besides, `attempts` will always be greater than zero except the first try, so it may be better to remove the conditional and print the message unconditionally.

philiplantz: I think this is intentional. If the first attempt works, no message is printed. If the first attempt fails, it prints a message saying it is retrying, and then prints a message if it is successful.

luisremis: @philiplantz that is correct. Usually there is no problem, so we don't show anything. If there is a problem, we try 10 times and print.

tests/python/TestCommand.py
```diff
@@ -41,6 +41,26 @@ def __init__(self, *args, **kwargs):
...
+                    print("Connection to VDMS successful.")
+            except:
+                print("Attempt", attempts,
+                      "to connect to VDMS failed, retying...")
+                attempts += 1
+                time.sleep(1) # sleeps 1 second
```
vishakha041: Does this commit belong in this branch or is this showing up cause GitHub is messing it up.

src/SearchExpression.cc
vishakha041: Skipping tests is fine but please separate out the commits. It makes it easier to debug when something doesn't work right.

tests/python/TestCommand.py
luisremis: are you asking this commit to be in a separate PR? We can do that, sure. I will split the PR and go ahead with the merge.

PR #120, opened 2019-07-30
Do not push 'read' operation on Video constructor

mahircg:
Currently, in some of the constructors of Video class, a 'read' operation is
enqueued into the operation queue to ensure that the video is read before
any other operation is performed. This, however, forces a read operation on
on a video object even if there are no operations to apply (i.e.
_operations is empty).

This patch defers read operations until they are absolutely necessary.
This is done by storing the operations in a linked list (rather than a
vector), and adding a read operation to the head of the queue before
perform_operations() invokes each operation in the queue.

### DIFF
--- a/include/vcl/Video.h
+++ b/include/vcl/Video.h
@@ -31,6 +31,7 @@
 
 #pragma once
 
+#include <list>
 #include <memory> // For shared_ptr
 #include <string>
 
@@ -292,7 +293,7 @@ namespace VCL {
 
         Codec _codec; // (h.264, etc).
 
-        std::vector<std::shared_ptr<Operation>> _operations;
+        std::list<std::shared_ptr<Operation>> _operations;
 
     /*  *********************** */
     /*        OPERATION         */
@@ -522,6 +523,14 @@ namespace VCL {
          *  @param Video_id  The full path to the Video to be read
          */
         void read(const std::string &video_id );
+
+        /**
+         *  Checks whether the video pointed by the current video_id has
+         *  already been read.
+         *
+         * @return true if video was read, false otherwise
+         */
+        bool is_read(void);
 
         /**
          *  Performs the set of operations that have been requested
--- a/include/vcl/Video.h
+++ b/include/vcl/Video.h
@@ -517,14 +517,6 @@ namespace VCL {
     /*       UTILITIES          */
     /*  *********************** */
         /**
-         *  Stores a Read Operation in the list of operations
-         *    to perform
-         *
-         *  @param Video_id  The full path to the Video to be read
-         */
-        void read(const std::string &video_id );
-
-        /**
          *  Checks whether the video pointed by the current video_id has
          *  already been read.
          *
--- a/src/vcl/Video.cc
+++ b/src/vcl/Video.cc
@@ -49,10 +49,10 @@ Video::Video() :
 {
 }
 
-Video::Video(const std::string& video_id)
+Video::Video(const std::string& video_id) :
+    Video()
 {
     _video_id = video_id;
-    read(_video_id); // push the read operation
 }
 
 Video::Video(void* buffer, long size) :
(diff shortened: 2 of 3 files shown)


src/vcl/Video.cc
```diff
@@ -201,12 +199,22 @@ void Video::read(const std::string &video_id)
...
     {
-        for (int x = 0; x < _operations.size(); ++x) {
-            std::shared_ptr<Operation> op = _operations[x];
+        bool is_empty = _operations.empty();
+        auto first_op = !is_empty ? _operations.front() : NULL;
+        if (is_empty || (first_op && first_op->get_type() != READ)) {
+            if (!_video_id.empty() && !is_read())
+                _operations.push_front(std::make_shared<Read>());
+        }
+        for (auto op : _operations) {
```
luisremis: will auto& also work here?

src/vcl/Video.cc
```diff
@@ -201,12 +199,22 @@ void Video::read(const std::string &video_id)
...
 void Video::perform_operations()
 {
     try
     {
-        for (int x = 0; x < _operations.size(); ++x) {
-            std::shared_ptr<Operation> op = _operations[x];
+        bool is_empty = _operations.empty();
+        auto first_op = !is_empty ? _operations.front() : NULL;
+        if (is_empty || (first_op && first_op->get_type() != READ)) {
+            if (!_video_id.empty() && !is_read())
```
luisremis: if !_video_id.empty(), we should throw

src/vcl/Video.cc
```diff
@@ -201,12 +199,22 @@ void Video::read(const std::string &video_id)
...
+
 void Video::perform_operations()
 {
     try
     {
-        for (int x = 0; x < _operations.size(); ++x) {
-            std::shared_ptr<Operation> op = _operations[x];
+        bool is_empty = _operations.empty();
+        auto first_op = !is_empty ? _operations.front() : NULL;
+        if (is_empty || (first_op && first_op->get_type() != READ)) {
```
luisremis: this can be simplified as (it is safe as .empty is evaluated first):
if ( (_operations.empty() || _operations.front()->get_type() != READ) && !is_read()) ) {
            if (!_video_id.empty())
                  throw
           _operations.push_front(std::make_shared<Read>());

include/vcl/Video.h
```diff
@@ -523,6 +524,14 @@ namespace VCL {
          */
         void read(const std::string &video_id );
```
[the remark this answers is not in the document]
mahircg: Yes, will do.

src/vcl/Video.cc
mahircg: Yes, that should work fine.

src/vcl/Video.cc
mahircg: That's the case where the execution reaches this function through the default constructor, which I believe should never  happen (hence 'throw').

src/vcl/Video.cc
mahircg: Yes, looks way more clean.

src/vcl/Video.cc
mahircg: I suppose that condition would be  `if (video_id.empty()) { throw ... }` though.

mahircg:
I'll separate the UT and will add the patch to remove the 'read'.

src/vcl/Video.cc
philiplantz: const auto & ?

mahircg: Last patch iterates through the elements that way.

include/vcl/Video.h
```diff
@@ -516,12 +517,12 @@ namespace VCL {
     /*       UTILITIES          */
     /*  *********************** */
         /**
-         *  Stores a Read Operation in the list of operations
-         *    to perform
+         *  Checks whether the video pointed by the current video_id has
+         *  has already been read.
```
mahircg: duplicate \has\

PR #121, opened 2019-08-05
Parse key frames

mahircg:


include/vcl/Video.h
```diff
@@ -176,6 +177,16 @@ namespace VCL {
...
 
+        /**
+         *  Invokes key-frame generation on the video, if the video is encoded
+         *  with a H264 encoder. Index, and byte offset/length of each key
+         *  frame is stored. This operation is independent of other prior
+         *  visual operations that may have been applied.
+         *
+         * @return List of KeyFrame objects
+         */
+        KeyFrameList get_key_frame_list();
```
luisremis: maybe const KeyFrameList& here too?

tests/python/TestVideos.py
```diff
@@ -96,6 +96,29 @@ def test_addVideo(self):
...
+        video_params["extract_kf"] = True
+        video_params["properties"] = props
+        video_params["codec"]      = "h264"
+
+        query = {}
+        query["AddVideo"] = video_params
+
+        response, obj_array = db.query([query], [[video_blob]])
+
+        self.assertEqual(response[0]["AddVideo"]["status"], 0)
```
luisremis: it would be nice to query for VD:KF and check we are getting the right number of i frames, as part of the same test.

tests/unit_tests/Video_test.cc
```diff
@@ -659,4 +659,33 @@ TEST_F(VideoTest, CropWrite)
...
+}
+
+TEST_F(VideoTest, KeyFrameExtractionSuccess)
+{
+    try {
+        VCL::VideoTest video_data(_video_path_mp4_h264);
+
+        auto key_frame_list = video_data.get_key_frame_list();
+
+        ASSERT_FALSE(key_frame_list.empty());
```
luisremis: same here, it would be nice key_frame_list has the right number of keyframes for that video

utils/src/api_schema/api_schema.json
```diff
@@ -664,6 +664,7 @@
     "AddVideo": {
       "properties": {
         "_ref":       { "$ref": "#/definitions/refInt" },
+        "extract_kf": { "type": "boolean" },
```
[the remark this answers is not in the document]
mahircg: index_frames sounds good, I'll change it

include/vcl/KeyFrameParser.h
```diff
@@ -0,0 +1,74 @@
...
+extern "C"
+{
+#include <libavformat/avformat.h>
+}
+
+namespace VCL {
+
+    struct KeyFrame {
+        uint64_t idx;
+	    int64_t  base;
```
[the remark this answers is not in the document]
mahircg: Yes, I've been meaning to fix this.

vishakha041:
How will these be used? I didn't see changes for FindVideo. Is that planned for the next PR?

vishakha041:
Could you please update the wiki too, with the index_frames variable?

mahircg:
> How will these be used? I didn't see changes for FindVideo. Is that planned for the next PR?

This was planned for another PR, which I created today.

> Could you please update the wiki too, with the index_frames variable?

I'll do that.

PR #126, opened 2019-09-25
Add flag to optionally extract key-frame information on videos

luisremis:
There is currently an incompatibility issue because of video encoding/decoding libraries on the develop branch. Some interfaces are not available on default Ubuntu 16 libraries.

This patch adds a flag that has to be specifically used in order to compile keyframe extraction.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -16,6 +16,10 @@ AddOption('--prefix', dest='prefix',
                       metavar='DIR',
                       help='installation prefix')
 
+AddOption('--video-keyframes', action='append_const', dest='cflags',
+                      const='-DVIDEO_KEYFRAMES',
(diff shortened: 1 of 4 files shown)


SConstruct
```diff
@@ -16,6 +16,10 @@ AddOption('--prefix', dest='prefix',
                       metavar='DIR',
                       help='installation prefix')
 
+AddOption('--video-keyframes', action='append_const', dest='cflags',
+                      const='-DVIDEO_KEYFRAMES',
+                      help= 'Build KeyFrame Optimazations for Video (Ubuntu 18)')
```
philiplantz: Typo in Optimizations.
(Note there is also a (different) typo in the commit message.)

PR #129, opened 2019-10-09
Allow object creation from local files in server

mahircg:
This patch introduces 'from_local_file' parameter to AddVideo command,
that allows object creation from local video files in server instead of
blobs transferred from the client.

### DIFF
--- a/src/VideoCommand.cc
+++ b/src/VideoCommand.cc
@@ -135,7 +135,14 @@ int AddVideo::construct_protobuf(
     int node_ref = get_value<int>(cmd, "_ref",
                                   query.get_available_reference());
 
-    VCL::Video video((void*)blob.data(), blob.size());
+    const std::string from_server_file = get_value<std::string>(cmd,
+                                                        "from_server_file", "");
+    VCL::Video video;
+    if (from_server_file.empty())
+        video = VCL::Video((void*)blob.data(), blob.size());
+    else
+        video = VCL::Video(from_server_file);
+
 
     // Key frame extraction works on binary stream data, without encoding. We
     // check whether key-frame extraction is to be applied, and if so, we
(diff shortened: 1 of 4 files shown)


src/VideoCommand.cc
```diff
@@ -135,7 +135,14 @@ int AddVideo::construct_protobuf(
     int node_ref = get_value<int>(cmd, "_ref",
                                   query.get_available_reference());
 
-    VCL::Video video((void*)blob.data(), blob.size());
+    std::string from_local_file = get_value<std::string>(cmd,
```
luisremis: this can be const.

src/VideoCommand.cc
```diff
@@ -207,6 +214,12 @@ Json::Value AddVideo::construct_responses(
     return ret;
 }
 
+bool AddVideo::need_blob(const Json::Value& cmd)
+{
+    const Json::Value& add_video_cmd = cmd[_cmd_name];
+    return !(add_video_cmd.isMember("from_local_file"));
```
luisremis: from_local_file should be a define. But I am sure we are not using defined for other similar cases. We can probably do a future pass and make all the cases defines. for now it should be good.

utils/src/api_schema/api_schema.json
```diff
@@ -663,13 +663,14 @@
...
-        "_ref":         { "$ref": "#/definitions/refInt" },
-        "index_frames": { "type": "boolean" },
-        "codec":        { "$ref": "#/definitions/vidCodecString" },
-        "container":    { "$ref": "#/definitions/vidContainerString" },
-        "link":         { "$ref": "#/definitions/blockLink" },
-        "operations":   { "$ref": "#/definitions/blockVideoOperations" },
-        "properties":   { "type": "object" }
+        "_ref":             { "$ref": "#/definitions/refInt" },
+        "index_frames":     { "type": "boolean" },
+        "from_local_file" : { "type": "string"},
```
luisremis: "from_local_file", from the user perspective, seems like it is a local file in the client machine. Maybe the right name should be "from_server_file" or something of the kind. what do you think?

vishakha041:
The name of the parameter also ignores the possibility of it being some object. Maybe blob_path?

luisremis:
yeah but this is specifc for the AddVideo call. It has to be a video file.

utils/src/api_schema/api_schema.json
mahircg: I agree, I'll change it to 'from_server_file'.

src/VideoCommand.cc
mahircg: I agree with that. I guess a single patch that defines all the parameters would do it.

PR #137, opened 2021-07-30
fixing Dockerfile. Previous version did not move valijson to correct …

omp87:
…path for compling vdms


omp87:
Fixing Dockerfile because previous version did not copy valijson to appropriate directory.

Ragaad (approved):

PR #152, opened 2022-12-02
update zlib to 1.2.13

momerath42:
1.2.12 is vulnerable to CVE-2022-37434, so they broke the link


cwlacewe:
Hi @momerath42. Thank you for your input. We really appreciate it. We have already updated this dependency for our next release. Please look for v2.4.0 coming soon.

cwlacewe:
Fixed in #156

PR #205, opened 2024-08-08
Address code scanning issues with remote functions and vdms client

rv355:


remote_function/udf_server.py
```diff
@@ -1,15 +1,12 @@
...
 import json
 from datetime import datetime, timezone
 import os
 import sys
-from collections import defaultdict, deque
-import skvideo.io
-import imutils
 import uuid
 from zipfile import ZipFile
+from werkzeug.utils import secure_filename
```
ifadams, line 9: Is this 3rd party utility safe to include as a part of the merge? If its already part of the requirements, yes, but otherwise we might need additional updates to make sure we dont break anything, add new licensing requirements, etc.

ifadams (commented):
Only concern is if we have any additional 3rd party libs included to make sure that we're correctly scanned, no licensing issues, requirements docs/installs updated....

remote_function/udf_server.py
rv355: The utility is automatically installed with Flask and has the same licensing requirements. We have already included Flask in the requirements. The secure_filename is the recommended way of addressing filename security issues in Flask-based servers.

ifadams: Alright, as long as we're not adding additional licensing, scanning requirements at the moment I'm going to approve.

ifadams (approved):
No major issues to address that I see.

PR #208, opened 2024-08-15
175 descriptor set optimizations batch inserts

ifadams:
Ready for initial review, tests are going to continue to be refined and developed some, as well as some cleanup, but we can start discussing the implementation.


ifadams:
Needs rebasing.

PR #209, opened 2024-08-15
175 descriptor set optimizations batch inserts

ifadams:
Initial development done, still some refinement and test updates to be done, but ready for initial review.

### DIFF
--- a/src/DescriptorsCommand.cc
+++ b/src/DescriptorsCommand.cc
@@ -341,8 +341,9 @@ AddDescriptor::AddDescriptor() : DescriptorsCommand("AddDescriptor") {
   //_use_aws_storage = VDMSConfig::instance()->get_aws_flag();
 }
 
+//update to handle multiple descriptors at a go
 long AddDescriptor::insert_descriptor(const std::string &blob,
-                                      const std::string &set_path, int dim,
+                                      const std::string &set_path, int nr_desc,
                                       const std::string &label,
                                       Json::Value &error) {
   long id_first;
@@ -351,21 +352,23 @@ long AddDescriptor::insert_descriptor(const std::string &blob,
 
     VCL::DescriptorSet *desc_set = _dm->get_descriptors_handler(set_path);
 
-    if (blob.length() / 4 != dim) {
+    //TODO this check no longer applies, should move it elsewhere
+    /*if (blob.length() / 4 != dim) {
       std::cerr << "AddDescriptor::insert_descriptor: ";
       std::cerr << "Dimensions mismatch: ";
       std::cerr << blob.length() / 4 << " " << dim << std::endl;
       error["info"] = "Blob Dimensions Mismatch";
       return -1;
-    }
+    }*/
+
 
     if (!label.empty()) {
       long label_id = desc_set->get_label_id(label);
       long *label_ptr = &label_id;
-      id_first = desc_set->add((float *)blob.data(), 1, label_ptr);
+      id_first = desc_set->add((float *)blob.data(), nr_desc, label_ptr);
 
     } else {
-      id_first = desc_set->add((float *)blob.data(), 1);
+      id_first = desc_set->add((float *)blob.data(), nr_desc);
     }
 
     if (output_vcl_timing) {
@@ -422,12 +425,20 @@ int AddDescriptor::add_single_descriptor(PMGDQuery &query,
     std::string label = get_value<std::string>(cmd, "label", "None");
     props[VDMS_DESC_LABEL_PROP] = label;
 
-    int dimensions;
-    const std::string set_path = get_set_path(query, set_name, dimensions);
+    int dim;
+    const std::string set_path = get_set_path(query, set_name, dim);
 
     if (set_path.empty()) {
         error["info"] = "Set " + set_name + " not found";
         error["status"] = RSCommand::Error;
+        return -1;
+    }
+
+    if (blob.length() / 4 != dim) {
+        std::cerr << "AddDescriptor::insert_descriptor: ";
+        std::cerr << "Dimensions mismatch: ";
+        std::cerr << blob.length() / 4 << " " << dim << std::endl;
+        error["info"] = "Blob Dimensions Mismatch";
         return -1;
     }
 
@@ -438,8 +449,8 @@ int AddDescriptor::add_single_descriptor(PMGDQuery &query,
         retrieve_aws_descriptorSet(set_path);
     }
 
-    //TODO modify descriptor
-    long id = insert_descriptor(blob, set_path, dimensions, label, error);
+    //TODO modify insert descriptor to handle batches
+    long id = insert_descriptor(blob, set_path, 1, label, error);
 
     if (id < 0) {
         error["status"] = RSCommand::Error;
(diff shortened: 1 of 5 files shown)


src/QueryHandlerPMGD.cc
```diff
@@ -199,7 +199,8 @@ bool QueryHandlerPMGD::syntax_checker(const Json::Value &root,
 
 int QueryHandlerPMGD::parse_commands(const protobufs::queryMessage &proto_query,
                                      Json::Value &root) {
-  Json::Reader reader;
+    printf("Parse commands Checker\n");
```
ifadams, line 202: Debug prints need removal.

utils/src/api_schema/api_schema.json
```diff
@@ -744,7 +749,8 @@
         "label":      { "type": "string" },
         "_ref":       { "$ref": "#/definitions/refInt" },
         "link":       { "$ref": "#/definitions/blockLink" },
-        "properties": { "type": "object" }
+        "properties": { "type": "object" },
```
ifadams, line 752: Change name from properties list to batch properties.

tests/python/TestDescriptors.py
```diff
@@ -206,6 +206,119 @@ def test_addSetAndDescriptorsDimMismatch(self):
...
+        desc_find["set"] = "rightbatchsize"
+        desc_find["results"] = {"list":["batchprop"]}
+
+        query = {}
+        query["FindDescriptor"] = desc_find
+
+        trans = []
+        blobs = []
+        trans.append(query)
+        response, img_array = db.query(trans, blobs)
```
ifadams, line 315: Missing assert validation return.

src/DescriptorsCommand.cc
```diff
@@ -409,94 +412,249 @@ void AddDescriptor::retrieve_aws_descriptorSet(const std::string &set_path) {
...
+        if (cmd.isMember("link")) {
+            add_link(query, cmd["link"], node_ref, VDMS_DESC_EDGE_TAG);
+        }
+
+        Json::Value props_edge;
+        query.AddEdge(-1, set_ref, node_ref, VDMS_DESC_SET_EDGE_TAG, props_edge);
+    }
+
+
+    /* TODO example iteration over properties list
```
ifadams, line 623: remove cruft

ifadams:
@rolandoquesada Question for you: I'm assuming our "end to end" tests from python do not really exercise coverage outside of the client because they are running against the server?

src/DescriptorsCommand.cc
```diff
@@ -409,94 +412,240 @@ void AddDescriptor::retrieve_aws_descriptorSet(const std::string &set_path) {
...
-  }
+    // Note dimensionse are based on a 32 bit integer, hence the /4 math on size
+    // as the string blob is sized in 8 bit ints.
+    nr_expected_descs = prop_list.size();
+    expected_blb_size = nr_expected_descs * dimensions * 4;
+
+    //Verify length of input is matching expectations
+    if (blob.length() != expected_blb_size) {
+        std::cerr << "AddDescriptor::insert_descriptor: ";
+        std::cerr << "Expectected Blob Length Does Not Match Input ";
```
araghuna1, line 558: Typo "Expectected"

src/DescriptorsCommand.cc
```diff
@@ -409,94 +412,240 @@ void AddDescriptor::retrieve_aws_descriptorSet(const std::string &set_path) {
...
+        std::cerr << blob.length() << " != " << expected_blb_size << std::endl;
```
araghuna1, line 559: The 2 lengths will show as numbers in the output & might not be clear which is expected & which is the actual length. Might be better to add a tag to each output number to mark this.

src/DescriptorsCommand.cc
```diff
@@ -409,94 +412,240 @@ void AddDescriptor::retrieve_aws_descriptorSet(const std::string &set_path) {
...
-      std::cout << "Deleted " << n << " files or directories\n";
+        retrieve_aws_descriptorSet(set_path);
     }
 
-    return -1;
-  }
+    // Note dimensionse are based on a 32 bit integer, hence the /4 math on size
+    // as the string blob is sized in 8 bit ints.
+    nr_expected_descs = prop_list.size();
+    expected_blb_size = nr_expected_descs * dimensions * 4;
```
araghuna1, line 553: kudos for the comment explaining the magic number! Given its used in a couple of places, might be good to use a const int or #define so its easy to change it *when* we have to change the dimension size in the future.

src/DescriptorsCommand.cc
```diff
@@ -409,94 +412,240 @@ void AddDescriptor::retrieve_aws_descriptorSet(const std::string &set_path) {
...
 
-  int node_ref = get_value<int>(cmd, "_ref", query.get_available_reference());
+    if (id < 0) {
+        error["status"] = RSCommand::Error;
 
-  query.AddNode(node_ref, VDMS_DESC_TAG, props, Json::nullValue);
+        if (_use_aws_storage) {
+            // delete files in set_path
+            std::uintmax_t n = fs::remove_all(set_path);
+            std::cout << "Deleted " << n << " files or directories\n";
```
araghuna1, line 573: Why do we want to delete all existing files for the descriptor set when an insert of a new descriptor fails?

src/DescriptorsCommand.cc
```diff
@@ -409,94 +412,240 @@ void AddDescriptor::retrieve_aws_descriptorSet(const std::string &set_path) {
...
+    if (blob.length() != expected_blb_size) {
+        std::cerr << "AddDescriptor::insert_descriptor: ";
+        std::cerr << "Expectected Blob Length Does Not Match Input ";
+        std::cerr << blob.length() << " != " << expected_blb_size << std::endl;
+        error["info"] = "FV Input Length Mismatch";
+        return -1;
+    }
 
-  props[VDMS_DESC_ID_PROP] = Json::Int64(id);
+    //TODO modify insert descriptor to handle batches
```
araghuna1, line 564: nit: can remove the comment in the code now :-)

src/DescriptorsCommand.cc
ifadams: This is some weird hold-over logic from the iterative implementation. My understanding is because they're staging local files before popping them over to the S3 target. I'm hesitant about nuking this cause I don't have enough insight yet to exactly _why_ they took this approach.

src/DescriptorsCommand.cc
```diff
@@ -409,94 +412,240 @@ void AddDescriptor::retrieve_aws_descriptorSet(const std::string &set_path) {
...
+
+    int expected_blb_size;
+    int nr_expected_descs;
+    int dimensions;
+
+    //Extract set name
+    const Json::Value &cmd = jsoncmd[_cmd_name];
+    const std::string set_name = cmd["set"].asString();
+
+    Json::Value props = get_value<Json::Value>(cmd, "properties");
```
araghuna1, line 528: can an insertDescriptor query contain both 'properties' and 'batch_properties'?
If not, then some warning/error response when both are seen is probably needed.

If yes, then the 'properties' likely contains common properties to be applied for each descriptor in the batch. In this case, the props extracted here should be added into cur_props around line 586-587.

src/DescriptorsCommand.cc
```diff
@@ -409,94 +412,240 @@ void AddDescriptor::retrieve_aws_descriptorSet(const std::string &set_path) {
...
+        //constraints for getting set node to link to.
+        Json::Value constraints;
+        Json::Value name_arr;
+        name_arr.append("==");
+        name_arr.append(set_name);
+        constraints[VDMS_DESC_SET_NAME_PROP] = name_arr;
+
+        bool unique = true;
+
+        // Query set node-We only need to do this once, outside of the loop TODO MOVE
```
araghuna1, line 610: The comment in the code makes sense. Why get the reference to the same set per iteration, when the set we want to refer to is the same.
Moving the related code (L590-L613) out of the loop would simplify the operations in the loop to just an AddNode, add_link, AddEdge. (Probably will reduce the overall insert time too as we avoid the queries to get DescriptorSet reference for every element in the batch!)

src/DescriptorsCommand.cc
```diff
@@ -409,94 +412,240 @@ void AddDescriptor::retrieve_aws_descriptorSet(const std::string &set_path) {
...
+  Json::Value prop_list = get_value<Json::Value>(cmd, "batch_properties");
+  if(prop_list.size() == 0){
+      rc = add_single_descriptor(query, jsoncmd, blob, grp_id, error);
+  } else {
+      rc = add_descriptor_batch(query, jsoncmd, blob, grp_id, error);
   }
 
-  Json::Value props_edge;
-  query.AddEdge(-1, set_ref, node_ref, VDMS_DESC_SET_EDGE_TAG, props_edge);
+  if(rc < 0) error["status"] = RSCommand::Error;
```
araghuna1, line 644: More descriptive error please. we have the jsoncmd that failed. Any information to identify the failing command will be very helpful to a client.

tests/python/TestDescriptors.py
```diff
@@ -206,6 +206,117 @@ def test_addSetAndDescriptorsDimMismatch(self):
...
+        self.assertEqual(response[0]["FindDescriptor"]["returned"],10)
+
+        self.disconnect(db)
+
```
araghuna1, line 318: Do we need a FindDescriptor test that searches using a blob (i.e. k_neighbor search)? This would confirm that the VCL level batch insert works correctly

src/DescriptorsCommand.cc
ifadams: Good catch, technically yes it can, but properties is ignored in favor of batch properties: I can just remove that code here.

src/DescriptorsCommand.cc
ifadams: Yup, was a mistake there, even had the TODO (oops!) , testing this now before pushing change.

src/DescriptorsCommand.cc
```diff
@@ -409,94 +412,240 @@ void AddDescriptor::retrieve_aws_descriptorSet(const std::string &set_path) {
...
+        name_arr.append(set_name);
+        constraints[VDMS_DESC_SET_NAME_PROP] = name_arr;
+
+        bool unique = true;
+
+        // Query set node-We only need to do this once, outside of the loop TODO MOVE
+        query.QueryNode(set_ref, VDMS_DESC_SET_TAG, link, constraints, results,
+                        unique);
+
+        //note this implicitly means that every node of a batch uses the same link
```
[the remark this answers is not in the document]
ifadams: Technically yes, but it would make for a very messy client side query, as each link reference would be linked to a Find* call. 

Generally I like the idea of being able to support batches, and I feel like we're demonstrating its not too bad to implement.

src/DescriptorsCommand.cc
ifadams: Let me check this, the message may be set elsewhewre, which is distinct from the status code.

tests/python/TestDescriptors.py
ifadams: Good call, I can probably copy-paste an existing test here.

src/DescriptorsCommand.cc
ifadams: Okay, yeah, the actual error info string is set earlier in all the cases I found, so this should be okay, though I agree its a little weird. Ill see if I can refactor.

ifadams:
Alright folks, I think this is ready for a merge, however we have some nits around code coverage.

ifadams:
@araghuna1 @s-gobriel any final updates/review comments?

araghuna1 (approved):

s-gobriel:
Very good work @ifadams on the batch inserts. I wonder if you think it is valuable to add a couple of c++ unit tests (in addition to the python tests you included for end-to-end to testing). These unittests can target the functionality to add in batches using the different engines and retrieve and check that the values are correct. but otherwise, I think this is good.

ifadams:
> Very good work @ifadams on the batch inserts. I wonder if you think it is valuable to add a couple of c++ unit tests (in addition to the python tests you included for end-to-end to testing). These unittests can target the functionality to add in batches using the different engines and retrieve and check that the values are correct. but otherwise, I think this is good.

Hey @s-gobriel, if you have concern about this, we can add in additional tests now, or if we want to get it merged immediately we should create an issue to immediately develop these tests as a task next quarter. I might suggest the latter, but I'm okay with either decision.

s-gobriel:
I understand. Let's add these tests in the next quarter, especially that the python test is covering the functionality

s-gobriel (approved):

Issue #213, opened 2024-09-10
Workflow: Fix coverage report updates


cwlacewe:
**Describe the bug**
Currently, the updated coverage reports aren't being pushed if the coverage threshold is not satisfied (Results job fails) even if manually approved. This is shown in PR #209 which will affect coverage comparisons for upcoming PRs.

High chances of new PRs failing because comparing to old coverage values/reports.

**Expected behavior**
Even if comparison is below threshold, update coverage reports/values for current PR. This will involve removing `success` check for Update job because currently the job is completely skipped if this occurs and doesn't update anything.

**Screenshots**
![image](https://github.com/user-attachments/assets/3edefada-7be6-4634-85c0-1c4ae294d09c)

cwlacewe:
Fixed with #217

PR #215, opened 2024-09-10
removing unique constraint for link based descriptor queries

ifadams:
Quick fix here!

We had a requirement for a unique response internally from the database query, which for link-based FindDescriptor queries would obviously not be guaranteed.


ifadams:
@cwlacewe well this is weird. It was a one line change and coverage hiccupped like that?

cwlacewe:
@ifadams, I noticed this. It's from your previous PR which needed addition testing. I created an issue I'm working on now (#213)

cwlacewe (approved):

PR #236, opened 2024-10-17
Fix issue related to the installation of Python

rolandoquesada:
The current setup_vdms.sh script is not installing the new version of Python due to an issue related to when comparing the current version value to the requested version. Also it is needed to update some of the symbolic links to point to the new Python version directory


cwlacewe (dismissed):
Looks good! Thanks!

ifadams:
@cwlacewe @rolandoquesada is there any risk that this might cause problems for other programs relying on a predictable python version set in $PATH?

rolandoquesada:
> @cwlacewe @rolandoquesada is there any risk that this might cause problems for other programs relying on a predictable python version set in $PATH?
Hi.
Yeah, you're right. I just uploaded the changes to call to the "python3" version instead of "python". 
Also I modified the setup_vdms.sh script to install a newer version of Python when the current python3 version is older than the minimum required.
Please let me know if you have any other suggestions. Thank you.

ifadams:
> Please let me know if you have any other suggestions. Thank you

@rolandoquesada I think my bigger concern is that updating Python has the potential to break existing dependencies if we're not careful, unless this is all being done within the docker container or a virtual environment.

docker/base/Dockerfile
cwlacewe, line 1: These changes aren't necessary within Dockerfile because we use a virtual env which is automatically used

.github/scripts/Dockerfile.checkin
cwlacewe, line 1: These changes aren't necessary within Dockerfile because we use a virtual env which is automatically used

rolandoquesada:
> > Please let me know if you have any other suggestions. Thank you
> 
> @rolandoquesada I think my bigger concern is that updating Python has the potential to break existing dependencies if we're not careful, unless this is all being done within the docker container or a virtual environment.

Hi Ian, thanks for the feedback, @cwlacewe as I was fixing an issue related to the condition for installing 

> > Please let me know if you have any other suggestions. Thank you
> 
> @rolandoquesada I think my bigger concern is that updating Python has the potential to break existing dependencies if we're not careful, unless this is all being done within the docker container or a virtual environment.

Hi Ian, thanks for the feedback. If that's the case, I think I could remove the lines for updating the path to Python in /usr/bin. However, now I am wondering if we should keep the conditional that checks if we need to install a new version of Python (that condition was before my new changes). @cwlacewe I am wondering what you think about it?

The code will be as follows:
python_installation

.github/scripts/Dockerfile.checkin
rolandoquesada: Hi, no problem. I will remove them.

docker/base/Dockerfile
rolandoquesada: Hi, no problem. I will remove them. Thanks.

cwlacewe:
> > > Please let me know if you have any other suggestions. Thank you
> > 
> > 
> > @rolandoquesada I think my bigger concern is that updating Python has the potential to break existing dependencies if we're not careful, unless this is all being done within the docker container or a virtual environment.
> 
> Hi Ian, thanks for the feedback, @cwlacewe as I was fixing an issue related to the condition for installing
> 
> > > Please let me know if you have any other suggestions. Thank you
> > 
> > 
> > @rolandoquesada I think my bigger concern is that updating Python has the potential to break existing dependencies if we're not careful, unless this is all being done within the docker container or a virtual environment.
> 
> Hi Ian, thanks for the feedback. If that's the case, I think I could remove the lines for updating the path to Python in /usr/bin. However, now I am wondering if we should keep the conditional that checks if we need to install a new version of Python (that condition was before my new changes). @cwlacewe I am wondering what you think about it?
> 
> The code will be as follows: python_installation

@rolandoquesada I think we should check if 3.12 exists, if so, make it the default python3 version; otherwise install and do the same. 
@ifadams do you have concerns with this regarding making 3.12 the python3 default version? Do you think we should create a virtual env on the user's system instead of changing the default version?

cwlacewe:
> @cwlacewe I dont have a very concrete concern so much as my own experience that in bare-metal installs updating python can be challenging in unpredictable ways. Sometimes "just works" other times explodes. Seems to depend on vagaries of the host environment.
> 
> My thought might be that we focus on the automation for container build, and either need a prompt or have just a note that "You need to install python 3.12 for host installation" in the docs. Ill totally admit virtual environments are a mystery to me.
> 
> Thoughts?

That would work. We can remove the python installation from script but check whether 3.12+ is installed. If it isn't installed, we can exit script with note advising to install it.  If it is installed, we can use the newest version >= 3.12. How does that sound @ifadams @rolandoquesada?

ifadams:
> That would work. We can remove the python installation from script but check whether 3.12+ is installed. If it isn't installed, we can exit script with note advising to install it. If it is installed, we can use the newest version >= 3.12. How does that sound @ifadams @rolandoquesada?

That works for me. I may be exercising an overabundance of caution, FWIW.

rolandoquesada:
> > That would work. We can remove the python installation from script but check whether 3.12+ is installed. If it isn't installed, we can exit script with note advising to install it. If it is installed, we can use the newest version >= 3.12. How does that sound @ifadams @rolandoquesada?
> 
> That works for me. I may be exercising an overabundance of caution, FWIW.

That works for me as well, I am going to do it in that way. Thank you @ifadams @cwlacewe

cwlacewe (commented):
Looks good

ifadams (approved):
Looks good to me

PR #247, opened 2024-10-28
Fix for Issue #240

### LINKED ISSUE #240: The test case called "ImageTest.RemoteMetadata" passes the test even when the udf_server doesn't execute the request
**Describe the bug**
The test case called "ImageTest.RemoteMetadata" passes the test even when the udf_server.py doesn't run successfully

**To Reproduce**
1. Open the udf_server.py file located at tests/remote_function_test directory.
2. Edit the file and add an invalid module name to be imported.
For example add the line "import blahblah" in the first line of the file.
udf_server_issue_01

3. Add the following line to the same file, specifically in the first line of the image_api() function:
```python
print("This message should be displayed in the log file", file=sys.stderr)
``` 
udf_server_issue_02

4. Save that file.
5. Go to the tests directory.
6. Run the following command:
```shell
./run_tests.sh -n ImageTest.RemoteMetadata
``` 
7. Verify the test passed. However, an error message is displayed on the screen:
```
Remote server is not running.
``` 
udf_server_issue_03

8. Open the file called "tests/tests_remote_log.log".
9. Verify that there is an error message in that log saying "ModuleNotFoundError: No module named 'blahblah'".
udf_server_issue_04

10. Verify that the request was not executed by udf_server server by checking that it is not logged the message added in the step 3 ("This message should be displayed in the log file") in that log file.
11.  Verify that the test is passing even when it should fail due to the error mentioned in the step 9.

**Expected behavior**
The test should fail.

rv355:
closes #240


rolandoquesada (approved):
Approved. Thanks for doing the changes.

cwlacewe (approved):

PR #248, opened 2024-10-28
Fix for issue 237

### LINKED ISSUE #237: VDMS hangs when the test "ImageTest.ImageLoop" is run and the udf_server doesn't return a valid value in  "/image" POST due to an invalid response message
**Describe the bug**
VDMS keeps in an infinite loop when it doesn't receive a valid response from the udf_server "/image" POST service

**To Reproduce**
1. Open the udf_server.py file located at tests/remote_function_test directory.
2. Locate the "/image" POST service in that udf_server.py file
3. Comment or remove all the code of that function and add a new line which returns an empty string (return "")
image

4. Run the ImageTest tests with the following command in the tests directory
image

5. Verify the testing gets frozen/hangs when the test called "ImageTest.ImageLoop" is run
image
 
**Expected behavior**
The test called "ImageTest.ImageLoop" should fail and VDMS should handle correctly the invalid response from the udf_server.

rv355:
closes #237


cwlacewe (approved):

rolandoquesada (approved):
Approved. Thank you for fixing it

PR #265, opened 2025-02-19
Address vulnerabilities with dependencies and fix deprecated datetime command

cwlacewe:
Closes #262 
Closes #264 
Fix deprecation of `datetime.datetime.utcnow()`


tests/tls_test/prep_certs.py
```diff
@@ -70,16 +72,18 @@ def generate_signed_certificate(
 
     issuer = issuer_certificate.subject
 
+    current_time = datetime.datetime.now(datetime.timezone.utc).replace(tzinfo=None)
```
rolandoquesada, line 75: duplicated but it works for me

rolandoquesada (approved):
The changes look good to me. Thank you

tests/tls_test/prep_certs.py
cwlacewe: They're in separate functions

rolandoquesada: my mistake, you're right. :)

PR #272, opened 2025-04-02
Format Dockerfiles, convert wget to curl, remove duplicate 'option' i…

cwlacewe:
…n CMakeLists due to cmake warning and centralized in main CmakeLists


docker/README.md
```diff
@@ -1,30 +1,61 @@
...
+    ```bash
+    docker build --build-arg=http_proxy --build-arg=https_proxy -f docker/base/Dockerfile -t vdms:latest .
+    ```
 
-If you are building with Kubernetes client support, use:
+* Build image with Kubernetes client support:
+    ```bash
     cd base
-    Dockerfile | docker build -t vdms:latest --build-arg USE_K8S="ON" .
+    Dockerfile | docker build -t vdms:latest --build-arg USE_K8S="on" .
```
rv355, line 32: User needs to use "ON," or is it case-independent?

docker/README.md
```diff
@@ -1,30 +1,61 @@
...
+    ```
 
-To run the docker image as a container, include the --net flag.
-This flag is needed as the server will be accepting connections on the default VDMS port (55555).
+    or build from main VDMS directory using:
```
rv355, line 35: User needs to use "ON," or is it case-independent?

docker/README.md
cwlacewe: @rv355 It is changed to case insensitive now while previously it only accepted lowercase in Dockerfiles. I can revert this if suggested but both work now

rv355: We don't need to revert. This would be better.

docker/README.md
cwlacewe: Flag is case insensitive.
@rv355, Is everything else okay to merge into branch?

rv355: Yes, rest everything looks fine.

Issue #277, opened 2025-04-03
Update the README.md files for including the new layout for the UDF script files in remote_function and user_defined_operations folders


rolandoquesada:
**Describe the bug**
Update the README.md files for including the new layout for the UDF script files in remote_function and user_defined_operations folders
Update the functions in those remote_function and user_defined_operations folders to include the temporary directory parameter as optional

cwlacewe:
Closed by #278

PR #278, opened 2025-04-03
Update the README files and also add the default value for the temporary directory parameter.

rolandoquesada:
The changes done to the code apply for the issue: https://github.com/IntelLabs/vdms/issues/277


remote_function/README.md
```diff
@@ -28,8 +29,10 @@ def run(ipfilename, format, options):
 
 ## Setup
 1. Copy the `remote_function` directory on the machine you want to run the remote server. Can be run on any location, independent of where VDMS is running. However, the location should be reachable from the machine that is running VDMS. You can also use `sparse-checkout` to only retrieve the `remote_function` directory from the VDMS repo.
-2. Create the operation scripts as python scripts and place them in the `remote_function/functions` directory.
-4. Follow the following steps to run the remote on port <port_number>.
+2. Copy `resources` directory (located at the root of the repository) next to the `remote_function` directory.
```
cwlacewe, line 32: For clarification, when you say `next to` should the `resources`  directory be copied into the `remote_function` directory? So resulting directory would be `remote_function/resources`? If this is correct, I suggest changing it to:
```suggestion
2. Copy `resources` directory (located at the root of the repository) into the `remote_function` directory.
```

remote_function/README.md
```diff
@@ -73,7 +76,8 @@ The client query should contain the following three parameters:
 ## Detailed Instructions for new remote operation
 We now provide an example to add a new operation `cardetect` as a remote operation that would work with VDMS. The `cardetect` operation detects cars in an image and creates a rectangle around all cars. This operation requires a pretrained model available in the form of `xml` file online.
 
-1. Copy `remote_function` directory to your remote server machine. Say the address is `my.remote.server` and you copy the folder in the `home` directory. The folder structure you have now will look something like this;
+1. Copy `remote_function` directory to your remote server machine. Say the address is `my.remote.server` and you copy the folder in the `home` directory.
+2. Copy `resources` directory (located at the root of the repository) next to the `remote_function` directory. The folder structure you have now will look something like this;
```
cwlacewe, line 80: Same comment as `Setup` section

user_defined_operations/README.md
```diff
@@ -86,8 +86,8 @@ The client query should contain the following two parameters:
 
 ## Detailed Instructions for new UDF
 We now provide an example to add a new UDF `cardetect`. The `cardetect` operation detects cars in an image and creates a rectangle around all cars. This operation requires a pretrained model available in the form of `xml` file online.
-
-1. Copy `user_defined_operations` directory to anywhere you want but on the same server that is running VDMS. Say you copy the folder in the `home` directory. The folder structure you have now will look something like this;
+1. Copy `user_defined_operations` directory to anywhere you want but on the same server that is running VDMS. Say you copy the folder in the `home` directory.
+2. Copy `resources` directory (located at the root of the repository) next to the `user_defined_operations` directory. The folder structure you have now will look something like this;
```
cwlacewe, line 90:
```suggestion
2. Copy `resources` directory (located at the root of the repository) into the `user_defined_operations` directory. The folder structure you have now will look something like this;
```

remote_function/README.md
rolandoquesada: Hi Chaunté, the "resources" and "remote_function" directories should be siblings.
As they are siblings directories, could you please tell me if I should clarify that or if the use of "next" could confuse to the reader?
Thank you.

user_defined_operations/README.md
rolandoquesada: Hi Chaunté, the "resources" and "user_defined_operations" directories should be siblings.
As they are siblings directories, could you please tell me if I should clarify that or if the use of "next" could confuse to the reader?
Thank you.

remote_function/README.md
cwlacewe: @rolandoquesada, I'm was a little confused with the `next to` term. When you say sibling directories, they should be in same parent directory? So say we have `parent/remote_function`, then the resources path should be `parent/resources`? If so, I suggest:
```suggestion
2. Copy `resources` directory (located at the root of the repository) into same directory as the `remote_function` directory.
```
What do you think?

rv355: The suggestion by @cwlacewe sounds good. 

Otherwise, you can include the `resources` directory copy instruction in Step 1 if it still sounds confusing. So you say, "Copy the `remote_function` and `resources` directories on the machine you want to run the remote server. Both should be in the same parent directory. It can be on any server, independent of where VDMS is running. However, the server should be reachable from the machine that is running VDMS."

rv355 (dismissed):

cwlacewe:
@rolandoquesada I like @rv355's suggestion better

user_defined_operations/README.md
rolandoquesada: Hi @cwlacewe and @rv355, as Rohit and I discussed by chat, I modified the README files to explain that the resources directory should be copied into the corresponding remote_function or user_defined_operations directory. Also, I modified the steps so the user can follow that approach.
Please let me know if you see any issues or if you have any suggestions.
Thank you

cwlacewe (dismissed):

user_defined_operations/README.md
```diff
@@ -147,8 +146,9 @@ def run(settings, message, input_params):
    |__udf_local.py
 |__resources
```
rv355, line 147: Shouldn't the `resources` directory be one level down inside `user_defined_operations`?

rolandoquesada: Yeah, you're right. Thanks for noticing it.

cwlacewe (approved):

PR #292, opened 2025-05-28
initial implementation of filter library API and memory management for filters

### LINKED ISSUE #282: Fast Post-Filtering Library for Vector Search
**Is your feature request related to a problem? Please describe.**
A filtered vector search consists of a query vector and a filter for a specific field. Current design of VDMS uses the descriptor indices for doing the vector search (KNN) and the filtering is done using the graph store (PMGD, neo4j)

Fast post-filtering library introduces a new layer of filtering that can be used after the vector search and before hitting the graph store to narrow down the candidate list considerably that satisfy both KNN and filter (to avoid sending too many queries to the graph store), and as a result improve the filtered-search throughput 

**Describe the solution you'd like**
1- A new library with an new API to: 
(a) create "summaries" which are vectors that matches a certain attribute that the user can apply a fast filter on.
(b) query these "summaries" to return a set of K vectors that satisfy being K-nearest neighbors and satisfy the attribute for filtering

2- Integrate the new library with the KNN flow and graph store search

s-gobriel:
The initial implementation of the Filters Library, (1) Public API defined and (2) memory management to create and free tables are implemented. Integration with VDMS search and insert can start.  Rest of functionality in Filter add/delete/lookup is WIP.


keirafadams (dismissed):
Ill trust the logic of the the various filter implementations (I dont have the background to deep dive) but high level API looks good.

cwlacewe (dismissed):
Look's good. Let's keep an eye on whether cout messages in VBF_Filter are sufficient or whether we should convert to exceptions.

cwlacewe (approved):

PR #308, opened 2025-09-08
182 erroneous read only on server start

### LINKED ISSUE #182: Erroneous Read-Only Error on Server Start
Most of the time, even with a "fresh" install and database, on server start a message is printed to the terminal as follows: 

```
PMGD Response:
{
   "info" : "ReadOnly: ",
   "status" : -1
}

[
   {
      "FailedCommand" : "Transaction",
      "info" : "Failed PMGD Transaction: ReadOnly: ",
      "status" : -1
   }
]
```

This is mildly confusing, given the server otherwise appears to function as expected, with clients able to connect and interact.

This error arises due to the auto_delete query that runs as the server launches. A fresh database has no data/metadata so the query necessarily hiccups. While this is technically valid behavior it is confusing from an operator/client perspective.

### LINKED ISSUE #307: Unexpected query response for Find operations on empty/new PMGD db
**Describe the bug**
When starting a brand-new instance of VDMS and no PMGD db available. Running a query node/edge transaction (FindImage for example) will result in a query response with a ReadOnly exception error.

**To Reproduce**
1. Create a brand-new instance of VDMS
2. Perform a FindImage command with any constraints, it doesn't really matter at this point since the db is empty.
3. Your query response will contain a ReadOnly exception error.

**Expected behavior**
Even when the DB is empty, querying a node or edge that does not exists should result in a response with a `"No entities found"` message instead of an error.

**Screenshots**
Query:

Image

Reponse:

Image

**Additional context**
Expected response:

Image

StevenRojasC:
This PR fixes an unexpected query response containing an exception message when running a Query Edge/Node transaction on a new/empty PMGD db instance.

Closes #182 
Closes #307

### DIFF
--- a/src/PMGDQuery.cc
+++ b/src/PMGDQuery.cc
@@ -676,6 +676,8 @@ void PMGDQuery::QueryNode(int ref, const std::string &tag,
                           const Json::Value &constraints,
                           const Json::Value &results, bool unique,
                           bool intermediate_query) {
+  _readonly = false;
+
   PMGDCmd *cmdquery = new PMGDCmd();
   cmdquery->set_cmd_id(PMGDCmd::QueryNode);
   cmdquery->set_cmd_grp_id(_current_group_id);
@@ -713,6 +715,8 @@ void PMGDQuery::QueryEdge(int ref, int src_ref, int dest_ref,
                           const std::string &tag,
                           const Json::Value &constraints,
                           const Json::Value &results, bool unique) {
+  _readonly = false;
+
   PMGDCmd *cmdquery = new PMGDCmd();
   cmdquery->set_cmd_id(PMGDCmd::QueryEdge);
   cmdquery->set_cmd_grp_id(_current_group_id);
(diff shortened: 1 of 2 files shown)


StevenRojasC:
Since the code coverage was reduced by more than 0.1% with these changes, the following screenshot summarizes the line numbers that tests are covering (new one on the left, original on the right):
image

Here's some references to the code lines that are not been executed by the tests:
PMGDQuery.cc (89-92): https://github.com/IntelLabs/vdms/blob/6ce7c3a063e43c5d84afd464e46326e1465b126d/src/PMGDQuery.cc#L89
PMGDQueryHandler.cc (200-203): https://github.com/IntelLabs/vdms/blob/6ce7c3a063e43c5d84afd464e46326e1465b126d/src/PMGDQueryHandler.cc#L200
QueryHandlerPMGD.cc (368-383): https://github.com/IntelLabs/vdms/blob/6ce7c3a063e43c5d84afd464e46326e1465b126d/src/QueryHandlerPMGD.cc#L368

As you can notice, all of the code that is no longer being executed by the tests is error handling code. This code was being executed before due to the exception thrown at https://github.com/omp87/pmgd/blob/f52a0b3c2c661bb07ae3218214557165f72a0b22/src/TransactionImpl.h#L189, which is exactly what this PR tries to fix, even though we don't have control over the pmgd codebase. Since the problem is fixed and this exception is never being reached, we would expect the error handling code to never run.

image

This is how the output would look like when this error comes up.

cwlacewe (approved):

